PR #7, opened 2019-03-07
Sets up a new package for the {yourOktaDomain} change 

bretterer:
In this PR, there is a new package created to deal with editing the `okta-preview-domain` sections of the page to swap out for your domains based on login.okta.com.

This will only work from approved okta domains, so localhost will not provide the swap.

This code is based on the current live code through jekyll while removing jquery dependency

### DIFF
--- a/packages/@okta/vuepress-plugin-my-okta/clientRootMixin.js
+++ b/packages/@okta/vuepress-plugin-my-okta/clientRootMixin.js
@@ -1,0 +1,59 @@
+export default {
+  mounted() {
+    let iframe = document.createElement('iframe')
+    iframe.id = 'myOktaIFrame'
+    iframe.src = 'https://login.okta.com'
+    iframe.style = 'display:none'
+    document.body.appendChild(iframe)
+
+    window.addEventListener('message', this.receiveMessage, false);
+    document.addEventListener('visibilitychange', this.checkSuccess, false);
+
+    window.addEventListener("load", () => {
+      document.querySelector('.content').innerHTML = document.querySelector('.content').innerHTML.replace(/https:\/\/{yourOktaDomain}/gi, '<span class="okta-preview-domain">https://{yourOktaDomain}</span>')
+      this.getMyOktaAccounts()
+    })
+  },
+
+  methods: {
+    checkSuccess: () => {
+      if (typeof document.hidden !== "undefined") {
+        if (!document.hidden) {
+          if ( document.querySelector('.okta-preview-domain').innerHTML == 'https://{yourOktaDomain}' ) {
+            // We haven't replaced yet: try again.
+            this.getMyOktaAccounts();
+          } else {
+            // We've succeeded: don't try again in the future.
+            document.removeEventListener('visibilitychange', this.checkSuccess, false);
+          }
+        }
+      }
+    },
+
+    getMyOktaAccounts: () => {
+      document.querySelector('#myOktaIFrame').contentWindow.postMessage({messageType: 'get_accounts_json'}, 'https://login.okta.com')
+    },
+
+    receiveMessage: () => {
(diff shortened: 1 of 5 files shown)


packages/@okta/vuepress-plugin-my-okta/clientRootMixin.js
```diff
@@ -0,0 +1,59 @@
+export default {
+  mounted() {
+    let iframe = document.createElement('iframe')
```
swiftone: nit: this should be `const`

packages/@okta/vuepress-plugin-my-okta/clientRootMixin.js
```diff
@@ -0,0 +1,59 @@
...
+    iframe.id = 'myOktaIFrame'
+    iframe.src = 'https://login.okta.com'
+    iframe.style = 'display:none'
+    document.body.appendChild(iframe)
+
+    window.addEventListener('message', this.receiveMessage, false);
+    document.addEventListener('visibilitychange', this.checkSuccess, false);
+
+    window.addEventListener("load", () => {
```
swiftone: nit: Why double quotes here but single elsewhere?

packages/@okta/vuepress-plugin-my-okta/clientRootMixin.js
```diff
@@ -0,0 +1,59 @@
...
+      document.querySelector('.content').innerHTML = document.querySelector('.content').innerHTML.replace(/https:\/\/{yourOktaDomain}/gi, '<span class="okta-preview-domain">https://{yourOktaDomain}</span>')
+      this.getMyOktaAccounts()
+    })
+  },
+
+  methods: {
+    checkSuccess: () => {
+      if (typeof document.hidden !== "undefined") {
```
swiftone: nit: doublequotes

packages/@okta/vuepress-plugin-my-okta/clientRootMixin.js
```diff
@@ -0,0 +1,59 @@
...
+          } else {
+            // We've succeeded: don't try again in the future.
+            document.removeEventListener('visibilitychange', this.checkSuccess, false);
+          }
+        }
+      }
+    },
+
+    getMyOktaAccounts: () => {
+      document.querySelector('#myOktaIFrame').contentWindow.postMessage({messageType: 'get_accounts_json'}, 'https://login.okta.com')
```
swiftone: nit: semicolon

packages/@okta/vuepress-plugin-my-okta/clientRootMixin.js
```diff
@@ -0,0 +1,59 @@
...
+    },
+
+    receiveMessage: () => {
+      if (event.origin !== 'https://login.okta.com' || !event.data) {
+        return;
+      }
+
+      var accountsExist = event.data.length;
```
swiftone: nit: these `var` should be `const`

packages/@okta/vuepress-plugin-my-okta/clientRootMixin.js
swiftone: Actually, many lines in this file are missing semicolons.

bretterer: Do we need semicolons?    A lot of the sample code and vuepress code goes without... Isn't it the es6 way?

swiftone: summarizing our slack conversation for posterity (below is paraphrase):

Me: My rule has always been "If you can't describe the situation where ASI will screw you over (and there is one), then you shouldn't be relying on it".

You: This seems to be the VuePress convention, so it makes sense to follow that.

Me: (single tear silently rolls down my face)

PR #17, opened 2019-03-14
Creates redirects in conductor file from the redirects.json

bretterer:
This PR addresses 404's because of missing redirects that Jekyll was doing via its magical black box.

### DIFF
--- a/packages/@okta/migrate-from-jekyll/index.js
+++ b/packages/@okta/migrate-from-jekyll/index.js
@@ -221,6 +221,17 @@ function run() {
 
 
   fs.writeFileSync(docsRoot+'/.vuepress/redirects.json', JSON.stringify(redirects, null, 2))
+
+  try {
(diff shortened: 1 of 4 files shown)


packages/@okta/migrate-from-jekyll/index.js
```diff
@@ -221,6 +221,17 @@ function run() {
 
 
   fs.writeFileSync(docsRoot+'/.vuepress/redirects.json', JSON.stringify(redirects, null, 2))
+
+  try {
+    fs.writeFileSync(docsRoot + '/conductor.yml', 'redirects: \r\n')
+    //Build conductor redirects
+    redirects.forEach((redirect) => {
+      fs.appendFile(docsRoot + '/conductor.yml', ' - from: ' + redirect.path + '\r\n   to: ' + redirect.redirect + '\r\n')
```
swiftone: did you mean `appendFileSync`?  This won't catch errors to appends because they will be async.

PR #19, opened 2019-03-15
Update conductor spacing

bretterer:
Adds correct spacing for conductor file.

### DIFF
--- a/packages/@okta/migrate-from-jekyll/index.js
+++ b/packages/@okta/migrate-from-jekyll/index.js
@@ -225,7 +225,7 @@ function run() {
   fs.writeFileSync(docsRoot + '/conductor.yml', 'redirects: \r\n')
   //Build conductor redirects
   redirects.forEach((redirect) => {
-    fs.appendFileSync(docsRoot + '/conductor.yml', ' - from: ' + redirect.path + '\r\n   to: ' + redirect.redirect + '\r\n')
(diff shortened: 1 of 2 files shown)


packages/@okta/migrate-from-jekyll/index.js
```diff
@@ -225,7 +225,7 @@ function run() {
   fs.writeFileSync(docsRoot + '/conductor.yml', 'redirects: \r\n')
   //Build conductor redirects
   redirects.forEach((redirect) => {
-    fs.appendFileSync(docsRoot + '/conductor.yml', ' - from: ' + redirect.path + '\r\n   to: ' + redirect.redirect + '\r\n')
+    fs.appendFileSync(docsRoot + '/conductor.yml', '  - from: ' + redirect.path + '\r\n    to: ' + redirect.redirect + '\r\n')
```
swiftone: When spacing is important I prefer template literals:
```
fs.appendFileSync(`${docsRoot}/conductor.yml`, `  - from: ${redirect.path}\r\n    to: ${redirect.redirect}\r\n`)
```

Also, is the \r\n accurate?  wouldn't that be based on the current file system?

bretterer: This only needs to work for my machine during migration.   This will not run once we migrate.   single use script

PR #24, opened 2019-03-18
Create a domain admin warning component

bretterer:


bretterer:
We do.  It is in the blacklist of the migration script

PR #37, opened 2019-03-20
Updates strings for 404's

bretterer:


packages/@okta/migrate-from-jekyll/rules/alterMarkdownStrings.js
```diff
@@ -126,6 +126,30 @@ function alterMarkdownStrings(file) {
...
+    }
+
+    if( line.includes('[Error Codes](error_codes)')) {
+      line = line.replace('[Error Codes](error_codes)', '[Error Codes](/reference/error_codes/)')
+    }
+
+    if( line.includes('(/docs/guides/saml_guidance)')) {
+      line = line.replace('(/docs/guides/saml_guidance)', '(https://www.okta.com/integrate/documentation/saml/)')
+    }
+
```
swiftone: Probably irrelevant for this one-time task, but this pattern reduces the odds of typos:

```
const toReplace = { 
  '[Session API](sessions#create-session-with-session-token)': '[Session API](/docs/api/resources/sessions#create-session-with-session-token)', 
  '(/docs/guides/saml_guidance)': '(https://www.okta.com/integrate/documentation/saml/)',
};

Object.keys( toReplace ).forEach( text => if( line.includes(text) ) { line = line.replace(text, toReplace[text]); } );
```

bretterer: Good to know... Ill put it in my back pocket!

PR #59, opened 2019-03-21
note on browser redirect for logout and authorize endpoints

susanharper-okta:
note on browser redirect for logout and authorize endpoints

https://oktainc.atlassian.net/browse/OKTA-202702


nbarbettini:
Thanks @susanharper-okta! 👍

PR #60, opened 2019-03-21
Refactor selenium test scripts

wayneearl-okta:
- changed initial navigation URL for quickstarts
- added some descriptions for expect statements


tests/selenium/spec/quickstarts-spec.js
```diff
@@ -30,7 +30,7 @@ describe('quickstarts page default selections spec', () => {
   const quickstartsPage = new QuickStartsPage('/quickstart/');
 
   beforeEach(util.itHelper(async () => {
-    await quickstartsPage.navigate('/quickstart/');
+    await quickstartsPage.navigate('/documentation/');
```
bretterer: Why is this change here?  are we just avoiding being on `/quickstart/` to start each test?

tests/selenium/spec/code-layout-spec.js
```diff
@@ -11,7 +11,8 @@ describe('code page spec (java/spring/)', () => {
   const codePage = new CodePage('/code/java/spring/');
 
   beforeEach(util.itHelper(async () => {
-    await codePage.navigate();
+    await codePage.navigate('/code/java/spring/');
+    await codePage.refresh();
```
bretterer: Can you explain the need for a refresh before each test?

tests/selenium/spec/table-of-contents-spec.js
```diff
@@ -13,14 +13,14 @@ describe('table of contents navigation spec', () => {
...
+    await tocPage.resizeXLarge();  // At smaller sizes, table of contents is hidden
     await tocPage.navigate('/test_page/');
     await tocPage.refresh();
-    await tocPage.resizeXLarge();  // At smaller sizes, table of contents is hidden
   }));
 
   it('has basic table of contents in the test page', util.itHelper(async () => {
-    expect(await tocPage.getLevelOneItemText()).to.equal('Test Page');
-    expect(await tocPage.getLevelTwoItemsText()).to.include.members([
+    expect(await tocPage.getLevelOneItemText(), 'Level one item text').to.equal('Test Page');
```
bretterer: What is this doing?

tests/selenium/spec/quickstarts-spec.js
wayneearl-okta: Yes. It seems that the navigation redirects don't always work right if it thinks it's already on the right page.

tests/selenium/spec/code-layout-spec.js
wayneearl-okta: The refresh feels like a hack, but I noticed that there are a couple of things that seem to work better (more consistently). One is checking the current URL. The other seems to be finding elements or checking active link text. I noticed when I was running these tests that this one was consistently having to retry to pass.

tests/selenium/spec/table-of-contents-spec.js
wayneearl-okta: it's getting all of the matching (by css class) items and compares the text.

PR #61, opened 2019-03-21
Updates default node version and README

vijetmahabaleshwar-okta:


.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - "10"
```
bretterer: Can we leave version 10 in as well as an optional version

vijetmahabaleshwar-okta: @bretterer - If we leave 10 here, tests will run on both versions 8 and 10. I'm not sure if they run in parallel though (most likely they do). But there is a chance that travis build time is increased. Do we need it? Your thoughts, @robertjd ?

bretterer: Sounds good.  We can address later if we need

PR #62, opened 2019-03-21
Bug Fixes

bretterer:
These are all bug fixes that were reported by Santhosh.

Here is a screenshot of all the changes that were made.

Screen Shot 2019-03-22 at 10 31 12 AM


packages/@okta/vuepress-site/documentation/index.md
```diff
@@ -45,6 +45,9 @@ languages:
 sections:
```
motikagan-okta: @bretterer, in the old site, this page would get automatically updated: you just had to create a new page in the 'use cases' or 'API reference' areas, and the link would be added to this page.

nbarbettini: Changing that was intentional. I don't want pages to magically show up on the index anymore. We are going to curate them even more strongly in the future to make sure the most valuable stuff is shown on the home page.

packages/@okta/vuepress-theme-default/assets/css/okta/pages/code.scss
```diff
@@ -8,3 +8,16 @@ ul.code-list {
   padding-left: 0px;
 }
 
+
+.language-content a:not([class="header-anchor header-link"]) {
```
santhoshbalakrishnan: CC: @edburyenegren-okta

edburyenegren-okta: Instead of using `:not`, could we attach a class to the desired stylings instead?

Or, if the purpose is to target "normal" links, maybe `a:not([class="*"])`?

packages/@okta/vuepress-theme-default/assets/css/okta/pages/code.scss
```diff
@@ -8,3 +8,16 @@ ul.code-list {
...
+  display: block;
+}
+
+.language-content ul:not([class="language-tabs"]) {
```
edburyenegren-okta: Same comment about `:not` here.

Is there a way we can get around using `!important` below, or are these styles mainly "legacy" CSS?

packages/@okta/vuepress-theme-default/assets/css/okta/pages/code.scss
bretterer: @edburyenegren-okta I would love to do this, but for now, I can not wrap arbitrary elements through Markdown.  I am looking at solution to be able to do this, but for now, using `:not` in the css is necessary to not include the deep link anchor for headers.  `.language-content` is only used on code pages, so it will not affect anything else.

packages/@okta/vuepress-theme-default/assets/css/okta/pages/code.scss
bretterer: @edburyenegren-okta I would love to do this, but for now, I can not wrap arbitrary elements through Markdown. I am looking at solution to be able to do this, but for now, using :not in the css is necessary to not include the deep link anchor for headers. .language-content is only used on code pages, so it will not affect anything else.

I will see if we can remove `!important` from this.

nbarbettini:
Thanks for the annotated screenshot @bretterer, that is very helpful 👌

packages/@okta/vuepress-theme-default/components/Code.vue
```diff
@@ -6,6 +6,7 @@
 
     <!-- Begin Content -->
     <div class="PageContent-main language-content">
+      <Breadcrumb></Breadcrumb>
```
nbarbettini: I love how declarative these Vue components are! ❤️

PR #65, opened 2019-03-21
updated the create a password policy request, and then updated all th…

susanharper-okta:
updated the create a password policy request, and then updated all the button locations with the new collection id

https://oktainc.atlassian.net/browse/OKTA-195928


packages/@okta/vuepress-site/docs/api/postman/password-policy.json
```diff
@@ -70,8 +70,7 @@
...
     {
```
bretterer: It looks like this `{` may be in the wrong place?

nbarbettini: I think this is just the Github parser being goofy. This file is auto generated. If it works in postman, it's all good 😄

PR #68, opened 2019-03-22
Okta Expression Language - String Functions layout fix

dragosgaftoneanu-okta:
Added pre-column for rows that have only examples and output in order to fix the layout of the table and not overlap with the sidebar menu.


bretterer:
Thank you for the fix @dragosgaftoneanu-okta !

Issue #71, opened 2019-03-22
Reformat all tables


bretterer:
After migration, there are a bunch of poorly formatted tables in code.  Most of them still render, but someone should go through and re-format all tables in the markdown files to make them look standardized.

<img width="994" alt="Screen Shot 2019-03-22 at 2 14 35 PM" src="https://user-images.githubusercontent.com/1906920/54844274-dc8fac80-4cac-11e9-9fba-2688555670a0.png">

To help with this, you can use the intelliJ plugin https://plugins.jetbrains.com/plugin/7550-pipe-table-formatter

nbarbettini:
For those using VS Code, there are good extensions that will prettify markdown tables. I use this one: https://marketplace.visualstudio.com/items?itemName=yzhang.markdown-all-in-one

PR #72, opened 2019-03-22
Fixes reported broken tables

bretterer:
This fixes the reported broken tables.   

Additional work on tables needs to be done.  See #71

### DIFF
--- a/packages/@okta/vuepress-site/docs/api/resources/oauth-clients/index.md
+++ b/packages/@okta/vuepress-site/docs/api/resources/oauth-clients/index.md
@@ -33,9 +33,9 @@ Adds a new client application to your organization
 ##### Request Parameters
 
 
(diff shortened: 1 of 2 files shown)


packages/@okta/vuepress-site/docs/api/resources/oauth-clients/index.md
```diff
@@ -276,7 +276,7 @@ Lists all clients that match a search filter on `client_name`
 
 
 Parameter | Description                                                                                | ParamType | DataType | Required | Default | Maximum |
```
swiftone: Why do the other tables start with a | but this one does not?

bretterer: Missed this one... thanks!

PR #73, opened 2019-03-22
feat(guides): Add WIP Guides layout and sample text

swiftone:


nbarbettini:
Quick question for Alex's input (don't worry @swiftone, I know this is WIP 😉):

1. The state of the page on first load - Java (the zeroth tab) is selected, but the URL does not contain "java".
image

2. The state of the page after clicking "js" - JS is selected, the URL contains "js"
image

Based on what we discussed before @alexdahl-okta, we eventually want:
1. If the reader has visited any step in this guide before, and picked a language, remember that choice for future visits to any step in this guide.
2. If the reader has never visited this guide, pick a default. (Is the zeroth tab a good default here? I'm assuming yes)
3. Regardless of remembered choice or default choice, the URL should reflect the choice value at all times.
Sound good to you @alexdahl-okta ?

swiftone:
> 1. If the reader has visited any step in this guide before, and picked a language, remember that choice for future visits to any step in this guide.

Define "future visits" - currently it updates the url and will remember it for all navigation within this guide (but not between guides).  Are you looking for some localStorage-style persistence? 

> 2. If the reader has never visited this guide, pick a default. (Is the zeroth tab a good default here? I'm assuming yes)

That's the default I'm using, but we can pick another (for example, there is meta data in the index.md of each guide - we can set a default-per-guide there if we want)

> 3. Regardless of remembered choice or default choice, the URL should reflect the choice value at all times.

Currently the url reflects an explicit choice, but if it's going off of default it won't (for example, if you explicitly click any tab, it will update the URL accordingly, and if you reload/revisit that url it will have that tab selected.

alexdahl-okta:
All good stuff – and yeah, not passionate that any of this has to be MVP. tl;dr sounds good @nbarbettini.

> 1. If the reader has visited any step in this guide before, and picked a language, remember that choice for future visits to any step in this guide.

Yes, I think we want some kind of local storage or cookie persistence (eventually).

And I think we could even be fancy across guides eventually... in my head:

- I imagine us saving a list of explicit user choices by order of recency. Say, `['java', 'android', 'go']`.
- Then when a user lands on a guide without an explicit choice, we try to find a match.
- Does this guide have Java as an option? If yes, switch to Java. If no, does it have Android as an option? Etc. If all fail, fallback to the zeroth tab.

> 2. If the reader has never visited this guide, pick a default. (Is the zeroth tab a good default here? I'm assuming yes)

I agree – extending "pick a default" to include the cross-guide persistence from above, if we have it.

If we can't identify a user preference, I do think the zeroth tab is a good default. I assume this will usually be alphabetical, but specifying the tab order would be one way to force a default.

> 3. Regardless of remembered choice or default choice, the URL should reflect the choice value at all times.

Yes, I think this would be nice to have. If the zeroth tab is Java, and that happens to be the choice the user wanted, reflecting Java in the URL may have a benefit:

- If the user bookmarks or shares the URL
- And later the zeroth tab changes – say, we add Go content as the new zeroth tab
- It would be nice for the URL to still bring them to Java

PR #75, opened 2019-03-22
Update old Liquid tag api_cors

nbarbettini:
Updates this:
image

to this:

image


nbarbettini:
For now, ApiLifecycle makes sense as a component of its own because it wraps up the logic for rendering different badges depending on `beta || ea || deprecated`.

Certainly could make sense down the road to have a more uniform way of displaying all info about an endpoint, like

```
<ApiEndpoint
  method="GET"
  route="/api/v1/users/${userId}"
  lifecycle="ga"
  supportsCors="true" />
```

We can tackle that later though IMO.

bretterer:
Fair Enough!

PR #81, opened 2019-03-23
Update installation instructions

nbarbettini:
Took a pass through the installation instructions, let me know if this direction makes sense.


README.md
```diff
@@ -13,47 +13,61 @@ The [Okta developer site][doc] serves Okta's API documentation and guides, inclu
...
+yarn dev
+```
 
-The directory structure from the `/packages/@okta/vuepress-site` folder relates directly to the URL you want to show. The `index.md` file that lives in the last folder in the structure will be rendered.
-If you name the file anything other than `index.md`, it will be required that you also include `.html` to the url when you go to view the page in the browser
+This starts a preview server on your machine, and watches all files for changes. Open http://localhost:8080/documentation/ to view the documentation.
 
-Part of the power that the VuePress development server gives us is live reloading of the content. When you make any changes to the markdown of the file, you will see the changes on the browser within seconds.
-However, if you make any chnages to the [front matter](https://github.com/vuejs/vuepress/blob/master/packages/docs/docs/guide/frontmatter.md) of the page, you may have to restart the development server.
+ > Note: if you try to visit the root, you will get a 404 page.  You must visit a path like `/documentation`.
```
bretterer: It is not any path.  it is only paths that are folders within the `packages/@okta/vuepress-site` folder

bretterer: It is not any path.  it is only paths that are folders within the `packages/@okta/vuepress-site` folder

README.md
```diff
@@ -13,47 +13,61 @@ The [Okta developer site][doc] serves Okta's API documentation and guides, inclu
...
+
+The preview server supports hot reloading. Once the server is running on your machine, any changes you make to Markdown files will appear automatically in your browser within a few seconds.
+
+## Adding and editing pages
+
+Documentation pages are stored as Markdown files in the `/packages/@okta/vuepress-site` directory.
+
+As an example, lets say you want to edit the [Users API](https://developer.okta.com/docs/api/resources/users) page. The public path of this page is `/docs/api/resources/users`.
```
bretterer: Should be `/docs/api/resources/users/` with the trailing slash

README.md
```diff
@@ -13,47 +13,61 @@ The [Okta developer site][doc] serves Okta's API documentation and guides, inclu
...
+To edit this page, you would navigate to `/packages/@okta/vuepress-site/docs/api/resources/users/index.md` and edit that Markdown file.
+
+An `index.md` file in a directory like `users` will be served as `/users` when the site is live. If you name the file anything other than `index.md`, you will need to include `.html` in the URL when you view the page in your browser.
```
bretterer: Served as `/users/`

README.md
```diff
@@ -123,6 +137,8 @@ A few options are provided for you to allow for some customization
 ### What About Building the Site Before Committing?
 There is no need to build the rendered site before committing and submitting a PR. This will all happen on the CI side to test and build the rendered site.
 
+## Running tests
+Running the tests before committing should be done and can be accomplished by running `yarn test` from the terminal. This will run a series of tests to make sure that everything is working as expected and that your changes did not affect anything that was not planned.
```
bretterer: we should add something like the following:

If you already have your local server running and available at `localhost:8080`, you should only run `yarn test-only` which bypasses the starting of the local server

At times, all of the tests will fail.  This is caused by the selenium suite not shutting down correctly. If this happens, run `yarn stop` and it should resolve itself on the next test run.

nbarbettini:
Fixed. Review again pls @bretterer?

Issue #48, opened 2019-03-21
Allow the use of site variables


bretterer:
During migration, we have to remove all `{{site.versions` for string replacement.  

We should allow for swapping these dynamically

Pages that were affected

- [ ]  packages/@okta/vuepress-site/code/angular/okta_angular_sign-in_widget/index.md
- [ ]  packages/@okta/vuepress-site/code/javascript/okta_auth_sdk/index.md
- [ ]  packages/@okta/vuepress-site/code/javascript/okta_sign-in_widget/index.md
- [ ]  packages/@okta/vuepress-site/quickstart-fragments/java/generic-implicit/index.md
- [ ]  packages/@okta/vuepress-site/quickstart-fragments/java/spring-auth-code/index.md
- [ ]  packages/@okta/vuepress-site/quickstart-fragments/java/spring-implicit/index.md
- [ ]  packages/@okta/vuepress-site/quickstart-fragments/widget/default-example/index.md
- [ ]  packages/@okta/vuepress-theme-default/global-components/ApiOperation.vue

bretterer:
After discussing, we're going to close this as a wont do. We want to spend more time investigating a better way of codifying the library versions in a yaml file, will follow up with a new velocity jira. For now we will manually update the URLs.

PR #88, opened 2019-03-25
Update signin-widget to 2.16

haishengwu-okta:


_config.yml
```diff
@@ -122,7 +122,7 @@ sitemap_ignore:
 versions:
   jwt_validator_java: 0.2.0
   spring_boot_starter: 0.5.0
-  okta_signin_widget: 2.6.0
+  okta_signin_widget: 2.16.0
```
bretterer: Hey!  Thanks for the update here, however at the moment `_config.yml` is not being used.  You will need to go to the each file to update the signin_widget version number.

You can use IntelliJ to search for the string `okta-signin-widget/2.14.0` to update all the required locations


This is a enhancement that we are going to be making soon to bring back this functionality. See: #48

haishengwu-okta: Thanks for reviewing @bretterer !
- update other files.
- keep changes in this config assuming it'll be used once the #48 has been solved.

Issue #108, opened 2019-03-26
Postman collections can't be accessed from developer.okta.com


dragosgaftoneanu-okta:
**Description:**
Postman collections under `/packages/@okta/vuepress-site/docs/api/postman` can not be accessed from developer.okta.com, returning 404.

**Example:**
https://developer.okta.com/docs/api/postman/example.oktapreview.com.environment

**Note:**
This collections can be accessed without any issues from okta.github.io, for example https://okta.github.io/docs/api/postman/example.oktapreview.com.environment

PR #109, opened 2019-03-26
Moves postman colleciton json to the public folder

bretterer:
It seems that non-markdown files are not being built during `yarn build` unless they are in the `.vuepress/public` directory.

This moves all postman json files over to the `.vuepress/public/docs/api/postman` folder


bretterer:
Resolves #108

PR #114, opened 2019-03-26
feat(guides): Adds Featured Guides

swiftone:


packages/@okta/vuepress-theme-default/util/frameworks.js
```diff
@@ -0,0 +1,21 @@
...
+};
+
+export const fancify = { 
+  javascript: 'JS',
+  angular: 'Angular',
+  react: 'React',
+  vue: 'Vue.js',
+  java: 'Java',
+  node: 'Node.js',
+};
```
bretterer: Would it make sense to defined these inside of the `.vuepress/config.js` file?

swiftone: I'm under the impression that that's for values about site configuration, not a dumping ground for all values, whereas this is more like a source file for an app - any part that interacts with frameworks can grab this file.  When we import it, we say WHAT we're importing, form a filename that informs.  

But that's pretty subjective.  Was this a question, or a polite "I'd prefer it to be here"?  I'm open to rethinking my position, certainly.

bretterer: It was a legit question.  That explanation tells me to keep it as is :)

PR #131, opened 2019-03-28
Guides update master

swiftone:


swiftone:
Closing because something isn't right

PR #133, opened 2019-03-28
Fixed the discrepancies in rate limits of shared buckets

prasanthmathialagan-okta:
Resolves: OKTA-213882

@okta/eng-perf 

Screen Shot 2019-03-28 at 1 19 03 PM


peteryaworsky-okta:
Rather than using the term "bucket" should we just say something like "shared with `/api/v1/authn`"?
What about `/app/office365/{key}/sso/wsfed/active` – the limit for `DEFAULT` is currently 1,000 I thought? Also why do the other SKUs not have values for `/app/office365/{key}/sso/wsfed/active` – does that mean they will map to `/` ?

jakubvul:
@prasanthmathialagan-okta is this targeted for a particular monolith release?

prasanthmathialagan-okta:
@jakubvul These are just bug fixes. I don't think they have to go with any particular release.

PR #144, opened 2019-03-29
Updates domain replacement

### LINKED ISSUE #137: {yourOktaDomain} swap not consistant
It appears that the update to fix #78 made it so not every instance of {yourOktaDomain} was swapped out.

bretterer:
There are cases where domain replacement string was not beginning with `https`.

This updates the replacement strategy to include `http://{yourOktaDomain}` to the replacement logic.

It also changes the logic to capture anything inside of `<code>` tags.  

Set up a section in the test_page that we can use to visually test the swap with all items that are capturing

resolves #137

### DIFF
--- a/packages/@okta/vuepress-plugin-my-okta/clientRootMixin.js
+++ b/packages/@okta/vuepress-plugin-my-okta/clientRootMixin.js
@@ -2,6 +2,10 @@ export default {
   mounted() {
     let iframeSrc = 'https://login.okta.com'
 
+    if(window.location.hostname == 'localhost') {
+      iframeSrc = 'https://login.okta.io:8081'
+    }
+
     const iframe = document.createElement('iframe')
     iframe.id = 'myOktaIFrame'
     iframe.src = iframeSrc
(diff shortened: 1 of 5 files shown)


packages/@okta/vuepress-plugin-my-okta/clientRootMixin.js
```diff
@@ -42,9 +46,10 @@ export default {
       });
       window.dispatchEvent(myOktaAccountFound)
 
-      const fences = document.querySelectorAll("[class*=language-]")
+      const fences = document.querySelectorAll("code")
       fences.forEach((fence, index) => {
         fences[index].innerHTML = fence.innerHTML.replace(/https:\/\/{yourOktaDomain}/gi, domain)
```
swiftone: Why not `fences[index].innerHTML = fence.innerHTML.replace(/https?:\/\/{yourOktaDomain}/gi, domain)` and skip doing it twice?

bretterer: Because `regex+bretterer=not smart`

PR #154, opened 2019-04-01
Guides seo

swiftone:
Updates the urls from `/guides/#sign-into-mobile/react/1` to `/guides/sign-into-mobile/react/1` for SEO purposes

This means we lose built-in behaviors: 
* everything scrolls to the top now (to be addressed in a scrolling PR)
* links default to loading a new page (links here are switched to use `<router-link to="/link">`)

On the up-side, we gain relative links between pages in guides (that will still load a new page unless we use `<router-link>`)


swiftone:
closing because of sync issues

PR #174, opened 2019-04-08
Guides seo sections

swiftone:
## Description:
- **What's changed?** 

Updates the Guides directory structure to support more SEO-friendly URLs.  As part of this, rewrote the logic that builds guides info from the vuepress page list - now, instead of re-parsing the array of pages for particular guide information, it builds a complete data structure on the first call and stores it in a frontend singleton.  Subsequent calls (on any guide/section, so long as you don't leave guides) will re-use this data.  Since we were looping through the entire pages array multiple times because of the number of details we needed even for a single section, this should reduce the frontend overhead.  Unfortunately there does not appear to be a way with vuepress to map these urls to generated HTML during SSR.

No

### Resolves:

* [OKTA-218133](https://oktainc.atlassian.net/browse/OKTA-218133)


README.md
```diff
@@ -47,7 +48,9 @@ As an example, lets say you want to edit the [Okta Angular Sign-in Widget](https
 To edit this page, you would navigate to `/packages/@okta/vuepress-site/code/angular/okta_angular_sign-in_widget/index.md` to edit the content of this page.
 
 The directory structure from the `/packages/@okta/vuepress-site` folder relates directly to the URL you want to show. The `index.md` file that lives in the last folder in the structure will be rendered.
-If you name the file anything other than `index.md`, it will be required that you also include `.html` to the url when you go to view the page in the browser
+If you name the file anything other than `index.md`, it will be required that you also include `.html` to the url when you go to view the page in the browser.
+
+The [Guides](https://developer.okta.com/guides/) section has a more complex directory structure.  See the "Updating Guides" section below.
```
laura-rodriguez: Should we add a link to the "Updating Guides" section? I know the section is right below but it's still useful for lazy people like me that don't want to scroll :)

PR #188, opened 2019-04-10
Add entry for E0000068

mbegan:
Not sure where else this might present itself but the LDAPi references it in response to a failed attempt to connect using password concatenated with OTP

## Description:
- **What's changed?** Adds an entry for Error code E0000068 - This error code is encountered when an incorrect passcode/answer is provided during an authentication request.

I encountered this error message with the LDAP Interface.  I am not sure where else it might present itself.

### Resolves:

No open JIRA or GH Ticket, it resolves itself #188 i can open an issue on this project if required but it seems superfluous.


bretterer:
@mbegan please fill out the comments of the description for your PR

PR #198, opened 2019-04-10
Guides next section button

swiftone:
## Description:
- **What's changed?** 

* Adds `NextSectionLink` for navigation between sections in Guides.
* Updates guides content to include `NextSectionLink`s
* Updates README with information on using `NextSectionLink`

No

### Resolves:

* [OKTA-219916](https://oktainc.atlassian.net/browse/OKTA-219916)
Screen Shot 2019-04-10 at 3 43 25 PM

### DIFF
--- a/packages/@okta/vuepress-theme-default/enhanceApp.js
+++ b/packages/@okta/vuepress-theme-default/enhanceApp.js
@@ -1,5 +1,7 @@
 import StackSelector from './components/StackSelector';
+import NextSection from './components/NextSection';
 
 export default ({ Vue }) => {
   Vue.component('StackSelector', StackSelector); // Force global register
+  Vue.component('NextSection', NextSection); // Force global register
 };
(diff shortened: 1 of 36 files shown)


swiftone:
> next to other elements like the stack selector

I did a pass through, and it LOOKS like everything has some built-in space around it except the stack selector, so I added a very specific next-section+stack-selector (and vice-versa) addition of space.

packages/@okta/vuepress-theme-default/enhanceApp.js
```diff
@@ -1,5 +1,7 @@
 import StackSelector from './components/StackSelector';
+import NextSectionLink from './components/NextSectionLink';
 
 export default ({ Vue }) => {
   Vue.component('StackSelector', StackSelector); // Force global register
+  Vue.component('NextSectionLink', NextSectionLink); // Force global register
```
bretterer: I missed this last time, but we should move the components that need to be global into the `global-components` folder as Vuepress does this for us automagically.

swiftone: I tried that and it didn't seem to work, but I'll give it another go.  I've got another effort I'm in the midst of to deal with some SSR issues, so if you are cool with it I'll merge this as is and move these in that PR (which makes sense because enhanceApp only runs client-side, so I have to move these to get SSR working.

bretterer: Sounds fine to me.

nbarbettini:
@swiftone Fix looks good! 🚢

PR #204, opened 2019-04-11
Updates slack notifications

vijetmahabaleshwar-okta:
## Description:
- Slack integration wasn't working as intended
- Deleted the older integration and created a new one
- New integration is working as intended and failures are seen in slack channel 

### Resolves:


tests/selenium/spec/api-tags-spec.js
```diff
@@ -12,7 +12,7 @@ describe('API tags check spec', () => {
 
   it('shows the Beta lifecycle tags', util.itHelper(async () => {
     await docsPage.navigate(util.fixUrl('/docs/api/getting_started/releases-at-okta'));
-    expect(await docsPage.hasBetaTags(), 'expects Beta tag to be present on page').to.be.true;
+    expect(await docsPage.hasBetaTags(), 'expects Beta tag to be present on page').to.be.false;
```
vijetmahabaleshwar-okta: @wayneearl-okta - Any idea how this test passed? I was trying to get a failure 😕

bretterer: isnt this the opposite of what we want now?   we want the beta tag to show, which means to.be.true i would think

PR #228, opened 2019-04-17
combing validating token signature and manage key rotation sections i…

susanharper-okta:
## Description:
- **Combining validating token signature and manage key rotation sections into one** 

### Resolves:

* [OKTA-155740](https://oktainc.atlassian.net/browse/OKTA-155740)


susanharper-okta:
Need to add John Gronberg and Venkat Viswanathan

PR #233, opened 2019-04-18
fix consent request example

aaronpk:
## Description:
* this is a GET request made by a browser, so drop the `curl` example including sending custom headers and the SSWS key
* rephrased text to indicate this is opened in a browser
* replaced hardcoded `state` with placeholder like `nonce` to avoid encouraging people to hardcode state value


nbarbettini:
cc @johngronberg-okta

PR #239, opened 2019-04-19
fix regex

gabrielsroka:
## Description:
when using a string to create a regex, escape the `.` but don't escape `/`

this wasn't completed in the old repo, nor migrated over
https://github.com/okta/okta.github.io/pull/2588
see discussion there


nbarbettini:
Note (because I missed this nuance at first): these are regexes in strings, not JavaScript regex literals.

robertjd:
What is the problem that we are trying to solve?

gabrielsroka:
`.` is a wildcard, it Will match any character. If you want to match a
literal dot, then you have to use `\\.`

PR #251, opened 2019-04-23
Update response startIndex in pagination example

joechiarella:
## Description:
- **What's changed?** 
Update response startIndex in pagination example.  Request is made to `/Users?count=1&startIndex=1` but response shows `startIndex: 0`.  As SCIM uses 1-based indexing for both the request and response, the response should show `startIndex: 1`.


jakubvul:
Thanks very much @joechiarella !

PR #256, opened 2019-04-24
Update header levels in Activate Factor section

joshreesjones-okta:
## Description:
On the Factors page, the Activate Factors section (https://developer.okta.com/docs/api/resources/factors/#activate-factor) has a lot of "Request Parameters", "Response Parameters", "Request Example", and "Response Example" headers in the table of contents:

Screen Shot 2019-04-24 at 3 57 57 PM

I'm guessing we don't want those there (correct me if I'm wrong, though). I updated the header levels so that it seemed more like what we'd want to see.

I did not verify what the table of contents looks like after these changes because I couldn't build locally. I am assuming that the table of contents is generated from the header structure in the Markdown file, but if not, let me know and I'll update whatever I need to.

### Resolves:

N/A

@bretterer


bretterer:
@nbarbettini This is actually going to be a thing for ALL pages.  do we want to do a new PR to fix all others (ie. search and replace for these?) or add to this PR?

nbarbettini:
@bretterer Good catch. Let's get this merged and create a follow up task.

@joshreesjones-okta Thanks for your continued contribs!

nbarbettini:
Follow up task: https://github.com/okta/okta-developer-docs/issues/328

nbarbettini:
Built locally and confirmed it looks how it should:

image

👍 🚢

PR #275, opened 2019-04-30
Default registry, cafile

aarongranick-okta:
## Description:

Sets default npm registry and sets cafile to default value (undefined), for this project only.

### Resolves:

Developers may face issues installing behind corporate proxies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
     "packages/@okta/*"
   ],
   "scripts": {
+    "preinstall": "npm config delete cafile",
     "dev": "yarn workspace @okta/vuepress-site dev",
     "build": "yarn workspace @okta/vuepress-site build",
     "migrate": "yarn workspace @okta/migrate-from-jekyll migrate",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -4,6 +4,7 @@
     "packages/@okta/*"
   ],
   "scripts": {
+    "preinstall": "npm config delete cafile",
```
swiftone: What impact does this have to someone that has an existing .npmrc config outside of this module?   (does it change that .npmrc file?)

aarongranick-okta: It does not change any files. The setting seems to remain in effect (globally) possibly until I restart my machine. `npm config get cafile`

nbarbettini:
We need someone with strap to test this and make sure it doesn't mess anything up.

package.json
aarongranick-okta: Wrong about that. It definitely removes it from the file.

swiftone: In that case, I'm definitely not in favor of this change in a repo that non-okta users can clone.  Thanks for checking!

aarongranick-okta:
I have not found a way of overriding `cafile` without causing destruction to `~/.npmrc`

PR #304, opened 2019-05-07
Fix broken link to "SAML Guidance"

albertchen-okta:
## Description:
SAML guidance link has been moved and the existing link is broken. 

New link should be: https://www.okta.com/integrate/documentation/single-sign-on/

No

### Resolves:

* no JIRA, manually found and fixed this issue.


albertchen-okta:
Hm, actually maybe https://www.okta.com/integrate/documentation/single-sign-on/ is the intended article, since that used to be on https://developer.okta.com/docs/guides/saml_guidance and now redirects to /integrate

jakubvul:
That is correct, the link should go to https://www.okta.com/integrate/documentation/single-sign-on/. Also, please use the template for this and any future PRs.

PR #345, opened 2019-05-15
adding back in slash

susanharper-okta:
## Description:
- **adding back in slash for link going to client authentication method section** 

### DIFF
--- a/packages/@okta/vuepress-site/authentication-guide/implementing-authentication/password/index.md
+++ b/packages/@okta/vuepress-site/authentication-guide/implementing-authentication/password/index.md
@@ -39,7 +39,7 @@ curl --request POST \
   rlnU9oZU4qWGrhQSM%3Dyd&scope=openid'
 ```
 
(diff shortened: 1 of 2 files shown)


packages/@okta/vuepress-site/authentication-guide/implementing-authentication/password/index.md
```diff
@@ -39,7 +39,7 @@ curl --request POST \
   rlnU9oZU4qWGrhQSM%3Dyd&scope=openid'
 ```
 
-> Important: The call to the `/token` endpoint requires authentication. In this case, it is a Basic Auth digest of the Client ID and Secret. You can find the client ID and secret in your application's General tab. For more on Basic Auth, see [Client Authentication Methods](/docs/api/resources/oidc#client-authentication-methods).
+> Important: The call to the `/token` endpoint requires authentication. In this case, it is a Basic Auth digest of the Client ID and Secret. You can find the client ID and secret in your application's General tab. For more on Basic Auth, see [Client Authentication Methods](/docs/api/resources/oidc#/client-authentication-methods).
```
motikagan-okta: The hash and slash (hey, that rhymes!) are transposed.

susanharper-okta: Fixed

PR #359, opened 2019-05-17
Delete test_page

nbarbettini:
Any reason to keep this around?


nbarbettini:
Ah! Failed tests mean the page is used during tests? This page also show up in search results, can we use a different (real) page for tests?

cc @vijetmahabaleshwar-okta

bjr-okta:
This page is required as we have elements set up that we use across the site here to confirm styling and functionality.  the items on this page do not change which make it great for tests

PR #368, opened 2019-05-21
Update limit for listing group members endpoint

dragosgaftoneanu-okta:
## Description:
- **What's changed?** The maximum limit returned by  /api/v1/groups/${groupId}/users is 1 000, not 10 000. This PR fixes the documentation to mention the correct number.


jakubvul:
Confirmed with @kunalsinha-okta

PR #400, opened 2019-05-31
Remove documentation on deprecated/unsupported tokens api.

danielzeller-okta:
## Description:
- **What's changed?**
Removes the tokens page: https://developer.okta.com/docs/api/resources/tokens/
Redirects to the OAuth2 token guide if navigated directly to
No
### Resolves:

* [OKTA-219880](https://oktainc.atlassian.net/browse/OKTA-219880)


packages/@okta/vuepress-site/.vuepress/redirects.json
```diff
@@ -245,7 +245,11 @@
   },
   {
     "path": "/docs/getting_started/tokens.html",
-    "redirect": "/docs/api/resources/tokens/index.html"
+    "redirect": "/authentication-guide/tokens/"
```
swiftone: @bretterer - When should this file be updated vs conductor.yml?

larsjohansen-okta:
Is `okta:master` the right branch to merge this into?

danielzeller-okta:
Good question @larsjohansen-okta 😄

swiftone:
> Is `okta:master` the right branch to merge this into?

merges to master will immediately (after CI tests) be pushed to the prod site.

packages/@okta/vuepress-site/.vuepress/redirects.json
bretterer: the redirects.json is not effective anymore.  It would only work if we began using the 404 page provided by VuePress, or had a page existing in the repo that we wanted to redirect from, but we should just be using conductor.yml for every redirect

jakubvul:
@larsjohansen-okta / @danielzeller-okta since this isn't tied to a particular release, yes, it should be aimed at `master`.

Doc'd here: https://github.com/okta/okta-developer-docs/wiki/Contributing-to-the-Site#all-other-content

PR #414, opened 2019-06-04
OKTA-163552: Adds missing Logs API `AuthenticationContext.authenticationStep` docs

bobtiernay-okta:
## Description:
- **What's changed?** 
Adds missing Logs API `AuthenticationContext.authenticationStep` docs. Also fixes the casing on the `AuthenticationContext` example.
No.

image

image

### Resolves:

* [OKTA-163552](https://oktainc.atlassian.net/browse/OKTA-163552)

### DIFF
--- a/packages/@okta/vuepress-site/docs/api/resources/system_log/index.md
+++ b/packages/@okta/vuepress-site/docs/api/resources/system_log/index.md
@@ -380,13 +380,13 @@ The `authenticationContext` contains metadata about how the actor is authenticat
 
 ```json
 {
-    "AuthenticationProvider": "ACTIVE_DIRECTORY",
-    "AuthenticationStep": 0,
-    "CredentialProvider": null,
-    "CredentialType": "IWA",
-    "ExternalSessionId": "102N1EKyPFERROGvK9wizMAPQ",
-    "Interface": null,
-    "Issuer": null
+    "authenticationProvider": "ACTIVE_DIRECTORY",
+    "authenticationStep": 0,
+    "credentialProvider": null,
+    "credentialType": "IWA",
+    "externalSessionId": "102N1EKyPFERROGvK9wizMAPQ",
+    "interface": null,
+    "issuer": null
 }
 ```
 In such a case, one can recognize that the user used an IWA credential to authenticate against an Active Directory instance. All of the user's future generated events in this login session will share the same `externalSessionId`.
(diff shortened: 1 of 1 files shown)


packages/@okta/vuepress-site/docs/api/resources/system_log/index.md
```diff
@@ -396,6 +396,7 @@ Among other operations, this response object can be used to scan for suspicious
...
 | ----------             | --------------------------------------------------------------                                                                                                                                                                      | ---------------                 | -------- | --------- | --------- |
 | authenticationProvider | The system that proves the identity of an actor using the credentials provided to it                                                                                                                                                | String                          | TRUE     |           |           |
+| authenticationStep     | The zero-based step number in the authentication pipeline (`0` - primary factor authn, `1` - secondary factor authn, etc).                                                                                                          | Integer                         | TRUE     |           |           |
```
AjayThomas-okta: Looks like this is only being used in an event currently fired by [DeviceTrust](https://github.com/okta/okta-core/blob/master/components/platform/device-trust/core/src/main/java/com/okta/monolith/platform/devicetrust/events/DeviceTrustEventHelperImpl.java#L339)

Also, looks like Passwordless might use it in the future which is owned by our team MFAPlatform

Description looks good to me
cc'ing other stakeholders

@mohammadrahimi-okta @sshen-okta @srinivasanagandla-okta

srinivasanagandla-okta: I wasn't aware of `authenticationStep`. As Ajay it seems to be used by device trust and I see all the usages have the value as `0`. 

I think we should remove reference to it for now and doc it when we are ready at a product level.

Rest of it looks good to me.

@tthompson-okta FYI...

bobtiernay-okta: Well, it shows up in the API response and the existing examples show it so for completeness I tend to think we should document it, but say that it is currently unused and always `0`. When the behavior changes we should then correct the docs.

As an aside, which is the correct team to handle the behaviour change in the platform?

bobtiernay-okta:
Made some updates to the description so that it is both accurate and complete:

image

packages/@okta/vuepress-site/docs/api/resources/system_log/index.md
srinivasanagandla-okta: @bobtiernay-okta I guess it is @okta/idx team. @tthompson-okta does that sound right to you?

I am still not sure if we want to document `authenticationStep` given that it could change.

srinivasanagandla-okta: 🚢

PR #424, opened 2019-06-06
Java self signed jwt guide

bdemers:


bdemers:
@nbarbettini I'm guessing you are looking for a little more content here based on the initial stubs of the guide.  

But I wanted to throw this out there to get the conversation started, other than a better intro link about the library used below, what content are we looking for?

nbarbettini:
👍 This is exactly what we needed! Nothing more required code-wise.

Can you sanity check my latest commit to make sure the code still makes sense? I wanted to explicitly call out what variables need to be in place before the code will run.

nbarbettini:
Then, merge into `sdh-okta221719-buildJWT` 🚢

PR #425, opened 2019-06-07
Minor updates to Custom URL Domain guide

nbarbettini:
I followed the Custom URL Domain guide for my own org and made a few changes to the doc while I was waiting for DNS to propagate. 😜


packages/@okta/vuepress-site/guides/custom-url-domain/enable-the-custom-domain/index.md
```diff
@@ -39,19 +39,16 @@ Okta serves traffic over HTTPS (TLS) on your custom domain. Use this section to
...
 
-    - `org.Subdomain.customdomains.okta.com`
-    - `org.Subdomain.customdomains.oktapreview.com`
-    - `org.Subdomain.customdomains.okta-emea.com`
-    
-2. Return to your Domain Name registrar and locate the option to modify your DNS records.
-3. Create a CNAME record and paste the value that you copied from the **Host** column into the appropriate field, for example, the **Name** or **Host** field. 
-4. Paste the value that you copied from the **Data** column into the appropriate field, for example, the **Record** or **Value** field.
-5. Save the record.
+1. Paste the value from the **Data** column into your registrar's **Record** or **Value** field.
```
susanharper-okta: Will it always be a **Record** or **Value** field at their registrar's? Do we know for sure? If so, I'm okay leaving out "for example", otherwise, "for example" implies that it MAY be called **Record** or **Value**, but it MAY be called something else. Your thoughts?

packages/@okta/vuepress-site/guides/custom-url-domain/enable-the-custom-domain/index.md
```diff
@@ -39,19 +39,16 @@ Okta serves traffic over HTTPS (TLS) on your custom domain. Use this section to
 
 ### Create a CNAME Record for Your Subdomain
 Before Okta can serve traffic over your domain, you need to add an alias from your custom domain to the Okta subdomain of your Okta organization. You do this by creating or modifying a CNAME record for your custom domain name.
+    
+1. Return to your Domain Name registrar and locate the option to modify your DNS records.
+1. Create a new CNAME record and paste the value from the **Host** column into your registrar's **Name** or **Host** field.
```
susanharper-okta: Will it always be a Record or Value field at their registrar's? Do we know for sure? If so, I'm okay leaving out "for example", otherwise, "for example" implies that it MAY be called Record or Value, but it MAY be called something else. Your thoughts?

packages/@okta/vuepress-site/guides/custom-url-domain/enable-the-custom-domain/index.md
nbarbettini: It MAY be called something else, yes. Different registrars have different UI so I'm not certain it will always be called that.

susanharper-okta: So, how about this:

Paste the value from the **Data** column into the appropriate field at the registrar, for example, the **Record** or **Value** field.

nbarbettini:
Addressed feedback and fixed up post-migration. This is ready to merge with your review @susanharper-okta. Please 🚢 !

PR #465, opened 2019-06-14
Sentence case for titles and headings, removed comments and added lin…

susanharper-okta:
…ks to guides that now exist

## Description:
- **Sentence case for titles and headings in guides, removed comments and added links to guides that now exist** 


nbarbettini:
IIRC the InfoDev team adopted sentence case from Design Systems. Can you check with Valli on a final decision here? My vote is for sentence case.

PR #504, opened 2019-06-27
Fixed missing TOC items

nbarbettini:
* Fixed: Rate Limits and Release Lifecycle links were not showing up
* Added some comments to help make it clear which files to change for the TOC and /docs landing page

Before:

missing-rate-limits

After:

added-rate-limits


nbarbettini:
I don't know why these didn't work as sublinks. I tried to fix that but failed. I'd prefer them as top-level links anyway, so I just moved them.

PR #506, opened 2019-06-27
created define callback section

susanharper-okta:
## Description:
- **created define callback section** 

### Resolves:

* [OKTA-234092](https://oktainc.atlassian.net/browse/OKTA-234092)


packages/@okta/vuepress-site/docs/guides/sign-users-out/define-signout-callback/aspnet/addbaseuri.md
```diff
@@ -0,0 +1 @@
+Follow your base URI with `/account/postsignout` (for example, `http://localhost:3000/account/postsignout`).
```
laura-rodriguez: ASP.NET 4.x is a particular case. In this case, users can define the `postLogout` url as they want, for example, `http://localhost:3000/foo/getmeoutofhere`. This is technically not a signout callback, but an "after-sign-out" callback. We have to clarify that.

laura-rodriguez: I just looked again, and it makes sense when you follow the entire guide. When you read the step-by-step instructions it is clear that this is a customizable URI.

PR #533, opened 2019-07-08
Remove outdated React Native quickstart content

robertjd:
## Description:
- **What's changed?** Removing content that will no longer work with the current version of the react native SDK. We will replace this with new guide content once the sample application is updated.


jakubvul:
@robertjd unfortunately we have a test that relies on that content being there to pass.

PR #536, opened 2019-07-09
Fix font family

alexdahl-okta:
## Description:

**What's changed?** Fixes font family back to Proxima Nova.
**Is this PR related to a Monolith release?** No.

- Okta's design system specifies Proxima with a different family name than we are loading, so it was not used despite its presence in the font stack.
- This changes our import of the design system from a single statement into 1) default abstracts 2) customized variables 3) the rest of the system. Today this is very verbose; eventually the system will make this easier.
- I've named the file handling this `odyssey.scss`. Okta's design system has been renamed to Odyssey, so this is forward looking. But the package on npm is still named `nim`, so we're in an awkward transition.
- There should be no other style changes.

proxima


alexdahl-okta:
@bretterer GitHub isn't letting me request a review from you, but I would like to!

packages/@okta/vuepress-theme-default/assets/css/okta.scss
```diff
@@ -2,7 +2,7 @@
 
 @import
 
-  '~@okta/nim/src/scss/nim',
```
bretterer: Why is it that we are getting rid of importing all of nim and instead individual items?

alexdahl-okta: Alas, because we need to insert a variable definition in the middle of nim.

If we define our variable before, nim's definition overrides it. If we define it after, we're too late and it will never apply.

This could be avoided if the design system declared variables with `!default`: https://github.com/okta/odyssey/issues/37

(The design system doesn't yet have fast velocity and regular releases, so I think we should work around it on our side for now.)

PR #558, opened 2019-07-23
Update docs to allow app group deletion

joshreesjones-okta:
## Description:
- **What's changed?** We now allow app group deletion for app groups not used in group push mappings, so this change removes the limitation specified in the docs.
- **Is this PR related to a Monolith release?** Yes, release 2019.07.2

### Resolves:

* [OKTA-214275](https://oktainc.atlassian.net/browse/OKTA-214275)

### DIFF
--- a/packages/@okta/vuepress-site/docs/reference/api/groups/index.md
+++ b/packages/@okta/vuepress-site/docs/reference/api/groups/index.md
@@ -849,10 +849,9 @@ curl -v -X PUT \
 
 <ApiOperation method="delete" url="/api/v1/groups/${groupId}" />
 
-Removes a group with `OKTA_GROUP` type from your organization.
+Removes a group with `OKTA_GROUP` or `APP_GROUP` type from your organization.
 
-> Only groups with `OKTA_GROUP` type can be removed.<br>
-> Application imports are responsible for removing groups with `APP_GROUP` type such as Active Directory groups.
+> Groups with type `APP_GROUP` cannot be removed if they are used in a group push mapping.
 
 ##### Request Parameters
 


packages/@okta/vuepress-site/docs/reference/api/groups/index.md
```diff
@@ -849,10 +849,9 @@ curl -v -X PUT \
 
 <ApiOperation method="delete" url="/api/v1/groups/${groupId}" />
 
-Removes a group with `OKTA_GROUP` type from your organization.
+Removes a group with `OKTA_GROUP` or `APP_GROUP` type from your organization.
```
motikagan-okta: There are no other possible group types except those two, right?  There's a chance this gives the impression that there are other types, that cannot be removed. Maybe:
"Removes a group from your organization . Groups of both types, `OKTA_GROUP` and `APP_GROUP`, can be removed. "

andreyshatniy-okta: We don't have more group types today, but we might in the future. So may be worth keeping this specific to Okta and App groups.

packages/@okta/vuepress-site/docs/reference/api/groups/index.md
```diff
@@ -849,10 +849,9 @@ curl -v -X PUT \
...
 
-> Only groups with `OKTA_GROUP` type can be removed.<br>
-> Application imports are responsible for removing groups with `APP_GROUP` type such as Active Directory groups.
+> Groups with type `APP_GROUP` cannot be removed if they are used in a group push mapping.
```
andreyshatniy-okta: Is group push the only check we're doing when deleting the app group? Anything else we should mention here?

joshreesjones-okta: To my knowledge, yes, this is the only check.

andreyshatniy-okta: Looks good to me then!

PR #584, opened 2019-08-02
Get Group notification update

dragosgaftoneanu:
## Description:
**What's changed?** 
- added note that the request to /api/v1/groups/{{groupId}} is supported only server side as confirmed by @AshishVaidya-okta on OKTA-226281
- corrected minor grammar in List Groups note

**Is this PR related to a Monolith release?**
No

### DIFF
--- a/packages/@okta/vuepress-site/docs/reference/api/groups/index.md
+++ b/packages/@okta/vuepress-site/docs/reference/api/groups/index.md
@@ -110,6 +110,8 @@ Fetches a specific group by `id` from your organization
 
 Fetched [Group](#group-model)
 
+> This request is not supported from a browser and should be called server side.
+
 ##### Request Example
 
 
@@ -185,7 +187,7 @@ Enumerates groups in your organization with pagination. A subset of groups can b
 | limit     | Specifies the number of group results in a page                                            | Query     | Number   | FALSE    | 10000   |
 | after     | Specifies the pagination cursor for the next page of groups                                | Query     | String   | FALSE    |         |
 
-> The `after` cursor should treated as an opaque value and obtained through the next link relation. See [Pagination](/docs/reference/api-overview/#pagination)
+> The `after` cursor should be treated as an opaque value and obtained through the next link relation. See [Pagination](/docs/reference/api-overview/#pagination)
 
 > Search currently performs a startsWith match but it should be considered an implementation detail and may change without notice in the future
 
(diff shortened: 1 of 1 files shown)


packages/@okta/vuepress-site/docs/reference/api/groups/index.md
```diff
@@ -110,6 +110,8 @@ Fetches a specific group by `id` from your organization
 
 Fetched [Group](#group-model)
 
+> This request is not supported from a browser and should be called server side.
```
jakubvul: I'm not clear on what this means.

AshishVaidya-okta: I think better wording would be:
`This request is not supported from a browser and should not be called from there.`

If we're willing to share the technical details, which I suspect devs would find out on their own anyway, then we should say something along the lines of:
`The API returns a JSON response and if a browser doesn't mention that in the 'Accept Headers' then the API shouldn't be called from there`

dragosgaftoneanu: Hi @jakubvul @AshishVaidya-okta 

Modified the note to `This request is not supported from a browser and should not be called from there.`.

dragosgaftoneanu: Hi @jakubvul @AshishVaidya-okta 

Any updates for this PR?

jakubvul: So is the problem essentially that you must include the `Accept: application/json` header in your request?

AshishVaidya-okta: Yes, that's my understanding following the comment from Brian Z. in the ticket. And that it happens for all our APIs. Are there reports of devs running into same problem for other APIs as the one here? Ideally, they should if browsers don't include `application/json` in the accept header. My understanding is that browsers don't request json by default -- https://developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation/List_of_default_Accept_values.

So if devs can somehow add that to the accept header while calling from browsers then things will work, just as they do when hitting our API thru Postman. We still need to fix the error response though. It should be 406, and not 5xx.

jakubvul: We tell people that they must include that header in our API overview page: https://developer.okta.com/docs/reference/api-overview/#media-types

It is also included in every sample request, including the one for this call.

dragosgaftoneanu: @jakubvul Removed the initial note and left only the grammar update.

PR #590, opened 2019-08-05
Clear current user sessions API doc

yumingcao-okta:
## Description:
- **What's changed?**
Screen Shot 2019-08-08 at 12 08 24 PM
Screen Shot 2019-08-08 at 12 08 52 PM

- **Is this PR related to a Monolith release?**  no - the monolith code has been released

### Resolves:

* [OKTA-234459](https://oktainc.atlassian.net/browse/OKTA-234459)

### DIFF
--- a/packages/@okta/vuepress-site/docs/reference/api/users/index.md
+++ b/packages/@okta/vuepress-site/docs/reference/api/users/index.md
@@ -2429,6 +2429,56 @@ HTTP/1.1 200 OK
 Content-Type: application/json
 ```
 
+### Clear Current User Sessions
+
+Clears Okta sessions for the currently logged in user. By default, the current session remains active. Use this method in a browser-based application. 
+
+> This operation requires a session cookie for the user. API token is not allowed for this operation.
+
+<ApiOperation method="post" url="/api/v1/users/me/lifecycle/delete_sessions" /> <SupportsCors />
+
+
+##### Request Parameters
+
+
+| Parameter    | Description                                                  | Param Type | DataType | Required | Default |
+| ------------ | ------------------------------------------------------------ | ---------- | -------- | -------- | ------- |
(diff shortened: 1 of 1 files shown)


packages/@okta/vuepress-site/docs/reference/api/users/index.md
```diff
@@ -2429,6 +2429,56 @@ HTTP/1.1 200 OK
 Content-Type: application/json
 ```
 
+### Clear Current User Sessions
+
+Clears all Okta sessions for the currently logged in user. This forces the user to authenticate on the next operation. Use this method in a browser-based application.
```
mauriciocastillosilva-okta: calling this API without any params will not force the user to authenticate again, since the default value for `keepCurrent` is true...
I think we should change it to be more explicit about it.. what about:
```
Clears Okta sessions for the currently logged in user. By default, the current session remains active. Use this method in a browser-based application. 
```

packages/@okta/vuepress-site/docs/reference/api/users/index.md
```diff
@@ -2429,6 +2429,56 @@ HTTP/1.1 200 OK
 Content-Type: application/json
 ```
 
+### Clear Current User Sessions
```
mauriciocastillosilva-okta: I like this title, so all good on my side
Can you double check with Damian as well please? Want to be sure it is clear that this is for current user, and that there is no confusion with the next topic right below this, which is for any user sessions (by admin or current user)

yumingcao-okta: @mauriciocastillosilva-okta yeah, I sync'ed with doc team, what do you think we move this section under `User Sessions` right below? So all session operations are doc'ed together?

mauriciocastillosilva-okta: talked to Yuming offline, we will keep it here since it is under `me/lifecycle`

PR #593, opened 2019-08-07
Updating index.md to refer to system API

kcbrewron:
I've added the warning from the event types page to the event hooks page. I was originally looking to use the event hooks to capture information using a webhook pattern. The changes to event types and system log API seem important enough to call out at the beginning of the Event Hooks documentation not a level in when looking at the event types

## Description:
- **What's changed?** 

### Resolves:

### DIFF
--- a/packages/@okta/vuepress-site/docs/reference/api/event-hooks/index.md
+++ b/packages/@okta/vuepress-site/docs/reference/api/event-hooks/index.md
@@ -16,6 +16,9 @@ For general information on event hooks and how to create and use them, see [Even
 
 Explore the Event Hooks API: [![Run in Postman](https://run.pstmn.io/button.svg)](https://app.getpostman.com/run-collection/b02d234a2af183981254)
 
+> **Important:** In the future the Events API will not be tracking new event types added to the System Log API. For this reason we highly recommend migrating to the System Log API.
+<br>
+
 ## Event Hook Operations
 
 ### Create Event Hook


packages/@okta/vuepress-site/docs/reference/api/event-hooks/index.md
```diff
@@ -16,6 +16,9 @@ For general information on event hooks and how to create and use them, see [Even
 
 Explore the Event Hooks API: [![Run in Postman](https://run.pstmn.io/button.svg)](https://app.getpostman.com/run-collection/b02d234a2af183981254)
 
+> **Important:** In the future the Events API will not be tracking new event types added to the System Log API. For this reason we highly recommend migrating to the System Log API.
+<br>
+
```
motikagan-okta: This page is about Event Hooks, not the Events API.

motikagan-okta: If I'm understanding correctly, I think you're saying that it would be helpful to mention that the event objects that are delivered by Event Hooks are the new kind of event objects used in System Log API, not the old kind of event objects used in the Events API.

That would probably belong on the Overview page for Event Hooks (https://developer.okta.com/docs/concepts/event-hooks/), rather than this page, which is just reference documentation for the management API.

kcbrewron: It is unclear if we can continue to implement okta event hooks to listen for user events. It seems that we should use the System Log API as any new events or event types will only be added to the System Log.

motikagan-okta: kcbrewron, just to confirm: you can certainly go ahead and implement Okta Event Hooks to listen for events. The events that are delivered by Event Hook are not being phased out or deprecated. The events delivered are Okta System Log API events, while the ones that are being phased out are Okta Event Log API events.

If there is confusion, I can add a clarification to the Event Hooks overview page saying that the events that are delivered are Okta System Log API events, not Okta Event Log API events.

Let me know if this makes sense.

motikagan-okta:
I've created a new pull request to add a note to the Event Hooks overview page, specifying that events delivered are the ones associated with the newer System Log API, not the ones associated with the older Events API:
https://github.com/okta/okta-developer-docs/pull/615
Going to close this pull request now.

PR #626, opened 2019-08-23
Removing anchors because they serve no purpose

jakubvul:
## Description:
- **What's changed?** I've removed the anchors from the Errors page because they don't do anything.


swiftone:
Neither should a ship rely on one small anchor, nor should life rest on a single hope.

- Epictetus

PR #680, opened 2019-09-17
Added missing enrollment and status attributes

AjayThomas-okta:
Added missing enrollment and status attributes for factorCatalogs when state is `MFA_ENROLL`

## Description:
- Enrollment attribute is a part of factorCatalog when authState is `MFA_ENROLL`. It is present if the org has or doe not have an MFA enrollment policy.
- Some responses in MFA_ENROLL state were missing `status` and `vendorName` attributes as well.

### Resolves:

* [OKTA-248739](https://oktainc.atlassian.net/browse/OKTA-248739)

### Reviewers:

@okta/mfaplatform-team


packages/@okta/vuepress-site/docs/reference/api/authn/index.md
```diff
@@ -461,6 +461,7 @@ User is assigned to a **MFA Policy** that requires enrollment during sign-on and
       {
         "factorType": "token",
         "provider": "RSA",
+        "vendorName": "RSA",
```
AjayThomas-okta: This could be RSA or a custom name mention in On Prem MFA settings.

AjayThomas-okta:
Another possible addition is adding `status` and `enrollment` attributes to the table defined in https://developer.okta.com/docs/reference/api/authn/#factor-object (Although these show only in `MFA_ENROLL` state)

packages/@okta/vuepress-site/docs/reference/api/authn/index.md
```diff
@@ -470,9 +471,11 @@ User is assigned to a **MFA Policy** that requires enrollment during sign-on and
               ]
             }
           }
-        }
+        },
+        "status": "NOT_SETUP",
+        "enrollment": "OPTIONAL"
       },
-      {
+      {
```
antonioruberto-okta: nit: extra whitespace added

jakubvul:
> Another possible addition is adding `status` and `enrollment` attributes to the table defined in https://developer.okta.com/docs/reference/api/authn/#factor-object (Although these show only in `MFA_ENROLL` state)

Well, if those properties are a part of the Factor object, then they should definitely be mentioned in that table.

AjayThomas-okta:
> LG2M - you tested to make sure JSON matches with an org that has OKTA_MFA_POLICY disabled correct?

Yes @antonioruberto-okta , I attached a screenshot to the [jira](https://oktainc.atlassian.net/browse/OKTA-248739)

AjayThomas-okta:
> > Another possible addition is adding `status` and `enrollment` attributes to the table defined in https://developer.okta.com/docs/reference/api/authn/#factor-object (Although these show only in `MFA_ENROLL` state)
> 
> @AjayThomas-okta If those properties are a part of the Factor object, then they should definitely be mentioned in that table.

I am not sure what `Factor Object` means in the doc.

- Status is part of both [UserFactor](https://github.com/okta/okta-core/blob/master/server/factors.interface/src/main/java/com/okta/services/factors/iface/dto/UserFactor.java) (which I think is what Factor Object is) and [FactorCatalogEntry](https://github.com/okta/okta-core/blob/master/server/factors.interface/src/main/java/com/okta/services/factors/iface/dto/FactorCatalogEntry.java).
- Enrollment is not part of UserFactor but is a part of FactorCatalogEntry

cc @antonioruberto-okta

PR #769, opened 2019-10-23
Fixes bug in sidebar and breadcrumb spacing

bretterer:
## Description:
- Breadcrumbs when displayed on old layout were showing an odd padding, this has been fixed with scoped css and props for breadcrumb (this fix is temporary until Prose layout is built
- Sidebar was not full due to an added div, this has been fixed

### DIFF
--- a/packages/@okta/vuepress-theme-default/components/Code.vue
+++ b/packages/@okta/vuepress-theme-default/components/Code.vue
@@ -5,9 +5,9 @@
     <!-- End Sidebar -->
 
     <!-- Begin Content -->
-    <div class="PageContent-main language-content">
+    <div class="PageContent-main language-content NoPaddingTop">
       <EditLink />
-      <Breadcrumb></Breadcrumb>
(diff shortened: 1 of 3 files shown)


packages/@okta/vuepress-theme-default/components/Code.vue
```diff
@@ -5,9 +5,9 @@
     <!-- End Sidebar -->
 
     <!-- Begin Content -->
-    <div class="PageContent-main language-content">
+    <div class="PageContent-main language-content NoPaddingTop">
       <EditLink />
-      <Breadcrumb></Breadcrumb>
+      <Breadcrumb wrapStyle="width: 80%"></Breadcrumb>
```
swiftone: Isn't there a better option than embedding style directly?

bretterer: I need the style directly on the .breadcrumb.Wrap, but only in this instance.   It's not the best, but will go away with the Prose redesign

PR #791, opened 2019-11-04
Fix a typo

nbarbettini:
As described.


nbarbettini:
@susanharper-okta My pleasure 🙂

PR #805, opened 2019-11-11
Prose style layout to concepts/references/release-notes

bretterer:
## Description:
- **What's changed?** Adds a temporary prose style layout to concepts/references/release-notes
- **Is this PR related to a Monolith release?** no

### release-notes
Screen Shot 2019-11-11 at 4 11 59 PM

### concepts
Screen Shot 2019-11-11 at 4 13 55 PM

### reference
Screen Shot 2019-11-11 at 4 14 03 PM

### Resolves:

* [OKTA-255232](https://oktainc.atlassian.net/browse/OKTA-255232)
* [OKTA-257095](https://oktainc.atlassian.net/browse/OKTA-257095)

### DIFF
--- a/packages/@okta/vuepress-site/.vuepress/nav/concepts.js
+++ b/packages/@okta/vuepress-site/.vuepress/nav/concepts.js
@@ -3,7 +3,8 @@
 module.exports = [
     {
         title: "Overview",
-        link: "/docs/concepts/"
+        link: "/docs/concepts/",
+        overview: true
     },
     { 
         title: 'API Access Management', 
--- a/packages/@okta/vuepress-site/docs/concepts/api-access-management/index.md
+++ b/packages/@okta/vuepress-site/docs/concepts/api-access-management/index.md
@@ -3,6 +3,7 @@ title: API Access Management
 meta:
   - name: description
     content: With API Access Management, you can secure all of your APIs. Read about its benefits and how to get started.
+layout: Landing
 ---
 
 # API Access Management with Okta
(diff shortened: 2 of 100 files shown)


bretterer:
> Looking purely at screenshots...
> 
> * 🐛 **Tree nav** appears to always be highlighting the first "Overview" item as if it were the active article. The active (currently-viewed) article should be highlighted with bolder type, blue background, and bottom border.

I have updated the tree nav to reflect figma.  The design choice here was knowing that we are rebuilding the whole theme to follow prose fully.  I copied design from existing guides for this.

> * 🐛 The **release notes tree nav** is including content from outside the Release Notes parent. I think it'll be fine if the tree nav for that section has only one `Release Notes` item. We can explore better long-term.

This was a debate on what to show here.  It looked odd to have release notes alone, but it also didn't make sense to have what I had for this PR.  I updated to have it as only `Release Notes` 

> * ❓ Are we causing any responsiveness regressions?

Its not the best UI, but it appears that current site does not even allow tree nav to show in mobile.  Right now, tree nav and content will show with scrolling to view content.  This is a temporary fix until the Prose layout is done, appx 4-6 weeks.  IMO, its a better UX than it is right now on the live site for mobile

> * ❓ Does a wider/fluid release notes width cause any problems?

No major issues here.  With `overflow: scroll` enabled temporally, it resolved the issue
> 
> Thanks!

bretterer:
Now allows for the breadcrumb `Show Contents` link to show and hide tree nav on small devices.

small devices will no longer show the tree nav to left of page

PR #829, opened 2019-11-27
fix minor error

gabrielsroka:
fix minor error, it's not `user.username`, it's `user.login`


jakubvul-okta:
Thanks @gabrielsroka !

PR #905, opened 2020-01-28
provide workaround for missing `ne` operator

gabrielsroka:
this is an FAQ, i thought it'd be useful to document here (or somewhere, if here is not appropriate)

feel free to verify and reword as necessary


jakubvul-okta:
Thanks @gabrielsroka and @colinwang-okta

PR #913, opened 2020-01-29
Change search to Coveo

hamzaessaoui-okta:
## Description:
- **What's changed?** 
Remove Swiftype and add Coveo Search

### Resolves:

* https://oktainc.atlassian.net/browse/BUSAPPS-17061


packages/@okta/vuepress-theme-default/assets/css/okta/base/layout.scss
```diff
@@ -6,10 +6,6 @@ body {
     background: get-color('white');
     position: relative;
 
-  @include media('<medium') {
```
bretterer: Why was this removed?

packages/@okta/vuepress-theme-default/components/TopNavigation.vue
```diff
@@ -132,7 +122,22 @@ export default {
     if(this.$themeConfig.search_url) {
       this.search_url = this.$themeConfig.search_url
     }
-  }
+  },
+
+  mounted() {
+    let coveoScript = document.createElement('script')
```
bretterer: Why is this happening here?  Can we just include the `okta_coveo_search_developer.js` file within the config.js like the other js and css files?

packages/@okta/vuepress-theme-default/assets/css/okta/base/layout.scss
benbridge: The top padding was pushing the search bar down below the header. With the top padding rule, the search bar was not hidden by default and then just moved down when the search icon was clicked resulting in this.

Screen Shot 2020-01-29 at 1-05-14-613-PM

packages/@okta/vuepress-theme-default/components/TopNavigation.vue
benbridge: This was to ensure that the entire Vue app is mounted before adding the script. Without this, some dom elements necessary for the Coveo framework initialization were missing and causing issues.

PR #931, opened 2020-02-04
Document new sortBy and sortOrder parameters

donwoods-okta:
## Description:
- Document new sortBy and sortOrder parameters
- The parameters apply only to "List Users" queries that use `search`
- Monolith release 2020.02.0

### Resolves:

* [OKTA-276081](https://oktainc.atlassian.net/browse/OKTA-276081)

### DIFF
--- a/packages/@okta/vuepress-site/docs/reference/api/users/index.md
+++ b/packages/@okta/vuepress-site/docs/reference/api/users/index.md
@@ -1374,6 +1374,7 @@ Use an ID lookup for records that you update to ensure your results contain the 
    - `sortBy` can be any single property, for example `sortBy=profile.lastName`
    - `sortOrder` is optional and defaults to ascending
    - `sortOrder` is ignored if `sortBy` is not present
+   - Users with the same value for the `sortBy` property will be ordered by `id`
    
 > **Note:** Searches that include a `sortBy` parameter may be slower.  As with other queries, even with pagination the results are limited to the 50000 users who come first according to the sort order.
 
--- a/packages/@okta/vuepress-site/docs/reference/api/users/index.md
+++ b/packages/@okta/vuepress-site/docs/reference/api/users/index.md
@@ -1361,7 +1361,7 @@ Property names in the search parameter are case sensitive, whereas operators (`e
 
 This operation:
 
-* Supports pagination (to a maximum of 50000 results; see note below).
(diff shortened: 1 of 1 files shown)


donwoods-okta:
@nghadge-okta please review for accuracy; @jakubvul please review for style and whatever magic is needed to ensure it goes live with 2020.02.0.

packages/@okta/vuepress-site/docs/reference/api/users/index.md
```diff
@@ -1368,6 +1370,12 @@ Use an ID lookup for records that you update to ensure your results contain the
    - Any user profile property, including custom-defined properties
    - The top-level properties `id`, `status`, `created`, `activated`, `statusChanged` and `lastUpdated`
    - The <ApiLifecycle access="ea" /> [User Type](/docs/reference/api/user-types), accessed as `type.id`
+* Accepts `sortBy` and `sortOrder` parameters.
+   - `sortBy` can be any single property, for example `sortBy=profile.lastName`
+   - `sortOrder` is optional and defaults to ascending
+   - `sortOrder` is ignored if `sortBy` is not present
+   
+> **Note:** Searches that include a `sortBy` parameter may be slower.  As with other queries, even with pagination the results are limited to the 50000 users who come first according to the sort order.
```
nghadge-okta: Where do we get this from? @jasonerickson-okta Do you see any such issue with `sortBy`? This is only applicable for search with `cursor=<Number>` eg `cursor=800`.

donwoods-okta: The existing doc says, specifically ONLY regarding search query:

Supports pagination (to a maximum of 50000 results; see note below).

But there isn't any note below that!  And now I'm told the limit only applies to customers who cheat by using a numeric cursor instead of the opaque cursor we tell them to use.  So I'm just going to delete that existing line and also remove it from my comment.  Now, what about "may be slower"?  I got that straight from the design wiki.

jasonerickson-okta: I don't see why a sorted query would be slower.  In fact, _all_ of the search results are always sorted - it's just a matter of getting to choose what they're sorted by.  Unlike an RDBMS, everything in ES is indexed, so you should be able to sort by whatever you like without noticeable performance impact.

PR #990, opened 2020-02-20
Case 00771223

delize:
This is creating an invalid search filter in Postman and other applications. Yes, it should be escaped, but a working example command should be used. Not an invalid one.

## Description:
- **What's changed?** 

### Resolves:


jakubvul-okta:
Hi @delize! Thank you for your PR, however after reviewing it, your change seems to not do what we would like to see.  Your PR states that an escaped value is correct, but you are removing the escape characters.  At this time, we are going to close the PR, but if you have any suggestions for another way to handle this, we welcome other PRs.

PR #1074, opened 2020-03-25
Correcting table

jakubvul-okta:
## Description:
- **What's changed?** Translations is an object, not an array.

### Resolves:

* [OKTA-286133](https://oktainc.atlassian.net/browse/OKTA-286133)

### DIFF
--- a/packages/@okta/vuepress-site/docs/reference/api/templates/index.md
+++ b/packages/@okta/vuepress-site/docs/reference/api/templates/index.md
@@ -371,13 +371,13 @@ All templates have the following properties:
 | template               | Text of the template, including any [macros](#sms-template-macros). | String (See note below)                                        | FALSE    | 1         | 161       |
 | created                | Timestamp when template was created                                 | String (ISO-8601)                                              | TRUE     | N/A       | N/A       |
 | lastUpdated            | Timestamp when template was last updated                            | String (ISO-8601)                                              | TRUE     | N/A       | N/A       |
-| translations           | Array of [translations](#translation-attributes)                    | Array                                                          | N/A      | N/A       | N/A       |
+| translations           | A list of [translations](#translation-attributes)                    | Object                                                          | N/A      | N/A       | N/A       |
 
 > **Note:** The final length of your SMS message cannot exceed 160 characters. If the verification code portion of the message falls outside of the 160-character limit, your message will not be sent.
 
 #### Translation Attributes
 
-Template translations are optionally provided when you want to localize the SMS messages. Translations are provided as key:value pairs: the language and translated template text.
+Template translations are optionally provided when you want to localize the SMS messages. Translations are provided as an object containing `key:value` pairs: the language and translated template text.
 
 ```json
 "translations": {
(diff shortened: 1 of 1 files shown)


packages/@okta/vuepress-site/docs/reference/api/templates/index.md
```diff
@@ -371,13 +371,13 @@ All templates have the following properties:
 | template               | Text of the template, including any [macros](#sms-template-macros). | String (See note below)                                        | FALSE    | 1         | 161       |
 | created                | Timestamp when template was created                                 | String (ISO-8601)                                              | TRUE     | N/A       | N/A       |
 | lastUpdated            | Timestamp when template was last updated                            | String (ISO-8601)                                              | TRUE     | N/A       | N/A       |
-| translations           | Array of [translations](#translation-attributes)                    | Array                                                          | N/A      | N/A       | N/A       |
+| translations           | A list of [translations](#translation-attributes)                    | Object                                                          | N/A      | N/A       | N/A       |
```
wdawson: Not sure this terminology is consistent with the rest of our API docs. See this [section on sign-on policy](https://developer.okta.com/docs/reference/api/policy/#signon-action-object) for an example of where I'm coming from.

- `List` is confusing to me as it implies an array, not a dictionary or key-value map. Perhaps something like `A key-value map for language translations based on a [user's language preferences](link-to-some-explanation-of-language-preferences)`
- The `Object` data type should be a `Translation Attributes` data type with a link to the section below.

jakubvul-okta: I will update the description, but the type should be one of the [JSON Data Types](https://www.w3schools.com/js/js_json_datatypes.asp).

packages/@okta/vuepress-site/docs/reference/api/templates/index.md
```diff
@@ -371,13 +371,13 @@ All templates have the following properties:
...
 | created                | Timestamp when template was created                                 | String (ISO-8601)                                              | TRUE     | N/A       | N/A       |
 | lastUpdated            | Timestamp when template was last updated                            | String (ISO-8601)                                              | TRUE     | N/A       | N/A       |
-| translations           | Array of [translations](#translation-attributes)                    | Array                                                          | N/A      | N/A       | N/A       |
+| translations           | A key-value map of [translations](#translation-attributes)                    | Translations Object                                                          | N/A      | N/A       | N/A       |
 
 > **Note:** The final length of your SMS message cannot exceed 160 characters. If the verification code portion of the message falls outside of the 160-character limit, your message will not be sent.
 
 #### Translation Attributes
```
thomasmaslen-okta: Is it a must for this heading to include the word "Attributes" (presumably for consistency with similar headings such as "SMS Attributes" above), or would it be OK for this heading to be just "Translations"?  In the table above we refer to this as "translations", and even though the href leads to the right place (good), it's still a little mental leap that ideally shouldn't be there.

jakubvul-okta: Our model is that, for every object, we have an "[Object] Properties" section that tells you what's inside that object.

PR #672, opened 2019-09-11
Create User Types API page based off Jakub's draft, tweaked per Googl…

donwoods-okta:
…e Doc

(Is it true that this PR and discussion will be visible publicly even though this is in my private fork??)

## Description:
- Create User Types API page based off Jakub's Beta draft, tweaked per Google Doc
- This is targeted for the October Monolith release; I don't know how to indicate that aside from this line.
- This change is currently only in my private fork of the public repo.

### Resolves:

* [OKTA-248765](https://oktainc.atlassian.net/browse/OKTA-248765)

PR #1076, opened 2020-03-27
DWW No more EA for User Types & fix some outdated text.  Delete User …

daytonwilliams-okta:
## Description:
- **What's changed?** User Types will be going GA in prod. 
- **Is this PR related to a Monolith release?** Yes, 2020.04.0

### Resolves:

* [OKTA-282464](https://oktainc.atlassian.net/browse/OKTA-282464)


packages/@okta/vuepress-site/docs/reference/api/users/index.md
```diff
@@ -1624,7 +1624,7 @@ in the request is deleted.
 
 `profile` and `credentials` can be updated independently or together with a single request.
 
->**Note:** Currently, the User Type of a user cannot be changed. If the Request Parameters include the `type` element from the [User object](#user-object), the value must match the existing type of the user. To change a User's type, the User object must be deleted and recreated with the desired Type.
+>**Note:** Currently, the User Type of a user can only be changed via a full replace. If the Request Parameters include the `type` element from the [User object](#user-object), on a partial update the value must match the existing type of the user. Only administrators are permitted to change the user type of a user; end users are not allowed to change their own user type.
```
colinwang-okta: Do we still need this note? I feel like it would be better to just specify under the partial update portion that type cannot be changed, and specify under the full update portion that it can be change.d

packages/@okta/vuepress-site/docs/reference/api/users/index.md
```diff
@@ -718,7 +718,7 @@ curl -v -X POST \
 }
 ```
 
->**Note:** The `type` property and the `schema` and `type` links will be present in all responses if the User Types feature is enabled, whether or not the user is created with a non-default User Type. See [User object](#user-object).
+>**Note:** The `type` property and the `schema` and `type` links will be present in all responses, whether or not the user is created with a non-default User Type. See [User object](#user-object).
```
colinwang-okta: Do we need this note anymore? I believe this was more to specify the behavior of flag on vs flag off, and that distinction no longer exists.

packages/@okta/vuepress-site/docs/reference/api/users/index.md
```diff
@@ -3789,7 +3787,7 @@ curl -v -X GET \
 }
 ```
 
->**Note:** The `type` element and the `schema` and `type` links are present only if the User Types feature is enabled. If enabled, they appear for all users, even those with the default User Type. See <ApiLifecycle access="ea" /> [User Types](/docs/reference/api/user-types).
+>**Note:** The `type` element and the `schema` and `type` links are present for all users, even those with the default User Type. See [User Types](/docs/reference/api/user-types).
```
colinwang-okta: Same question as above. In my opinion we might not need this anymore.

packages/@okta/vuepress-site/docs/reference/api/users/index.md
```diff
@@ -648,9 +648,9 @@ curl -v -X POST \
 
 #### Create User with Non-Default User Type
 
-Creates a user with a specified User Type (see <ApiLifecycle access="ea" /> [User Types](/docs/reference/api/user-types)). The type specification may be included with any of the above Create User operations; this example demonstrates creating a user without credentials.
+Creates a user with a specified User Type (see [User Types](/docs/reference/api/user-types)). The type specification may be included with any of the above Create User operations; this example demonstrates creating a user without credentials.
 
-The User Type determines which [Schema](/docs/reference/api/schemas) applies to that user. After a user has been created, the user currently cannot be assigned a different User Type.
+The User Type determines which [Schema](/docs/reference/api/schemas) applies to that user. After a user has been created, the user currently can only be assigned a different User Type by an administrator via a full replace update.
```
donwoods-okta: I would consider removing "currently", and also moving the word "only" to the part that it pertains to.  (I.e., many other things can be done to the user, but it can be assigned a different type only by blah blah blah.)  Finally, consider "using" instead of "via".

... the user can be assigned a different User Type only by an administrator using a full replace update.

packages/@okta/vuepress-site/docs/reference/api/users/index.md
donwoods-okta: 👍

packages/@okta/vuepress-site/docs/reference/api/users/index.md
donwoods-okta: Hm, I feel there's some value to the bit about how it's okay to include a type in a partial update but only if the type is not changing.

packages/@okta/vuepress-site/docs/reference/api/users/index.md
daytonwilliams-okta: I agree with all those except I like 'via' :) .  Changed.

packages/@okta/vuepress-site/docs/reference/api/users/index.md
daytonwilliams-okta: Yeah, I wondered about that, but ultimately decided the bit about 'schema' and 'type' links was marginally worth keeping.  I'm fine removing this.

packages/@okta/vuepress-site/docs/reference/api/users/index.md
daytonwilliams-okta: This is the only place in the whole API, as far as I know, where changing a thing (the user type) can only be done via a full update, and the partial update equivalent of the same API does not support it.  Calling that out explicitly seems prudent.  It is a fairly odd and unexpected thing to do.

packages/@okta/vuepress-site/docs/reference/api/users/index.md
daytonwilliams-okta: Yeah, I agree with the previous comment that this may as well come out.

packages/@okta/vuepress-site/docs/reference/api/users/index.md
donwoods-okta: I agree the full/partial distinction is worth keeping.  Nit: I think the second sentence flows a bit more smoothly, and makes the contrast with "full replace" more obvious, if the "partial update" part is moved earlier:

... via a full replace.  If the Request Parameters on a partial update include the `type` element from the ..., the value must ...

daytonwilliams-okta:
I did a few things:

1)  I searched for the string `[User Type` across all the documentation.  (This actually found a few references in the release notes that I left alone, since our release notes should continue to reflect that it went EA in the past).

2)  I looked at every occurrence of `<ApiLifecycle access="ea" />` in the files involved in this PR.

3)  I looked at every place in your original doc PR (here https://github.com/okta/okta-developer-docs/pull/672) where that string was added.

daytonwilliams-okta:
Also, picked your nit.  I agree, that is more readable.

packages/@okta/vuepress-site/docs/reference/api/users/index.md
```diff
@@ -648,9 +648,9 @@ curl -v -X POST \
 
 #### Create User with Non-Default User Type
 
-Creates a user with a specified User Type (see <ApiLifecycle access="ea" /> [User Types](/docs/reference/api/user-types)). The type specification may be included with any of the above Create User operations; this example demonstrates creating a user without credentials.
+Creates a user with a specified User Type (see [User Types](/docs/reference/api/user-types)). The type specification may be included with any of the above Create User operations; this example demonstrates creating a user without credentials.
 
-The User Type determines which [Schema](/docs/reference/api/schemas) applies to that user. After a user has been created, the user currently cannot be assigned a different User Type.
+The User Type determines which [Schema](/docs/reference/api/schemas) applies to that user. After a user has been created, the user can be assigned a different User Type only by an administrator via a full replace update.
```
jakubvul-okta, line 653:
```suggestion
The User Type determines which [Schema](/docs/reference/api/schemas) applies to that user. After a user has been created, the user can be assigned a different User Type only by an administrator via a full replacement [PUT operation](/docs/reference/api/user-types/#update-user-type).
```

packages/@okta/vuepress-site/docs/reference/api/users/index.md
```diff
@@ -1624,7 +1622,7 @@ in the request is deleted.
 
 `profile` and `credentials` can be updated independently or together with a single request.
 
->**Note:** Currently, the User Type of a user cannot be changed. If the Request Parameters include the `type` element from the [User object](#user-object), the value must match the existing type of the user. To change a User's type, the User object must be deleted and recreated with the desired Type.
+>**Note:** Currently, the User Type of a user can only be changed via a full replace. If the Request Parameters of a partial update include the `type` element from the [User object](#user-object), the value must match the existing type of the user. Only administrators are permitted to change the user type of a user; end users are not allowed to change their own user type.
```
jakubvul-okta, line 1625:
```suggestion
>**Note:** Currently, the User Type of a user can only be changed via a full replacement [PUT operation](/docs/reference/api/user-types/#update-user-type). If the Request Parameters of a partial update include the `type` element from the [User object](#user-object), the value must match the existing type of the user. Only administrators are permitted to change the user type of a user; end users are not allowed to change their own user type. 
```

packages/@okta/vuepress-site/docs/reference/api/users/index.md
```diff
@@ -3817,7 +3811,7 @@ Metadata properties such as `id`, `status`, timestamps, `_links`, and `_embedded
 * The `activated` timestamp will only be available for users activated after 06/30/2013.
 * The`statusChanged` and `lastLogin` timestamps will be missing for users created before 06/30/2013 and updated on next status change or login.
 
-The `type` property is a map that identifies the User Type of the user (see <ApiLifecycle access="ea" /> [User Types](/docs/reference/api/user-types)). Currently it contains a single element, `id`, as shown in the Example. It can be specified when creating a new User, but once the User is created the value is read-only.
+The `type` property is a map that identifies the User Type of the user (see [User Types](/docs/reference/api/user-types)). Currently it contains a single element, `id`, as shown in the Example. It can be specified when creating a new User, and may be updated by an administrator on a full replace of an existing user (but not a partial update).
```
jakubvul-okta, line 3814:
```suggestion
The `type` property is a map that identifies the User Type of the user (see [User Types](/docs/reference/api/user-types)). Currently it contains a single element, `id`, as shown in the Example. It can be specified when creating a new User, and may be updated by an administrator on a [full replace of an existing user](/docs/reference/api/user-types/#update-user-type) (but not a partial update).
```

donwoods-okta:
I like all of Jakub's suggestions.

packages/@okta/vuepress-site/docs/reference/api/users/index.md
daytonwilliams-okta: Done.

PR #1079, opened 2020-03-30
Guide for Password Import Inline Hook

motikagan-okta:
## Description:
- **What's changed?** Creating a new guide, to cover the coding of an external service for the Password Import Inline Hook.

### Resolves:

* [OKTA-280794](https://oktainc.atlassian.net/browse/OKTA-280794)


packages/@okta/vuepress-site/docs/guides/password-import-hook/check-authorization-header/nodejs/check-authorization-header.md
```diff
@@ -0,0 +1,14 @@
+For the check of the authorization in header here, we use Node packages to parse the request from Okta and to decode the Authorization header.
+
+We're going to choose to use the Basic Authentication scheme to encode a username and password combination in base64 to serve as the content of the Authorization header.
```
kenwang-okta: Should also mention Basic Authentication is used here for only for demonstration purposes and that our customers should use a stronger form of Authentication.

packages/@okta/vuepress-site/docs/guides/password-import-hook/web-server/nodejs/web-server.md
```diff
@@ -0,0 +1,5 @@
+```javascript
+const app = express();
+const port = process.env.PORT || 3000 ;
```
kenwang-okta: Should we put a note somewhere that this example uses HTTP but Okta will require HTTPS so such a server should be placed behind a secure reverse proxy?

packages/@okta/vuepress-site/docs/guides/password-import-hook/check-authorization-header/nodejs/check-authorization-header.md
motikagan-okta: As discussed, there is only support for setting the Authorization header to a static string.

packages/@okta/vuepress-site/docs/guides/password-import-hook/web-server/nodejs/web-server.md
motikagan-okta: As discussed, that's covered in password-import-hook/web-server/index.md: 
```
Inline Hook requests sent by Okta are HTTPS requests. To listen for them, your external service needs to implement an HTTPS server provisioned with a certificate from a trusted Certificate Authority.

This guide does not cover the specifics of this part of implementing an external service. In the following example code, we just listen for HTTP requests and assume that our code is running behind a reverse proxy configured for HTTPS.

>**Note:** Okta will not call a non-HTTPS external service endpoint. You must implement HTTPS in order to use Inline Hooks. 
```

PR #1083, opened 2020-03-31
Update index.md

gabrielsroka:
salt is not required (at least not for md5 -- i didn't try every type of hash...)

it should say "FALSE" or "only required for the salted hashes..." (like under saltOrder)


jakubvul-okta:
Fixed here: https://github.com/okta/okta-developer-docs/pull/1084/files

PR #1084, opened 2020-03-31
Make salt optional for import password hash object

wdawson:
The salt is optional when importing hashed passwords for users. Correct the doc which said it was required.

## Description:
- **What's changed?** Changed the salt requirement from `TRUE` to `FALSE` for the password hash import operation on create users
- **Is this PR related to a Monolith release?** No, this is a correction to the docs
### Resolves:

* N/A (no jira)

@rameshbhojan-okta can you clarify that this is accurate? I thought we implemented it this way.
@motikagan-okta FYI


rameshbhojan-okta:
Yes, the salt is optional, and that is how we implemented it to support customers who used the legacy hashes without salt. Also, see this desc. on the hash `value` which indicates the same. https://github.com/okta/okta-developer-docs/pull/1084/files#diff-30a1ece534324887cb721a01415994aaR3993

PR #1087, opened 2020-03-31
Sdh okta262819 logouri

susanharper-okta:
## Description:
- **What's changed?** Updated content around the logo_uri property

### Resolves:

* [OKTA-262819](https://oktainc.atlassian.net/browse/OKTA-262819)


packages/@okta/vuepress-site/docs/reference/api/apps/index.md
```diff
@@ -1137,7 +1137,7 @@ Adds an OAuth 2.0 client application. This application is only available to the
...
-| logo_uri                                    | URL string that references a logo for the client. This value is automatically copied from any logo you provide in the App Wizard for **Application logo**, but can be changed to a different URI for consent.               | URL                                                                                            | TRUE       | FALSE    | FALSE      |
+| logo_uri                                    | URL string that references a logo for the client.<br><br> This value is used with the client consent dialog box during the client consent flow when you:<ul><li> Create an app using the API and specify a value for `logo_uri`</li><li>Create an app in the UI and specify a value for **Application logo**.</li></ul>This value is used as the logo on the application's tile for the dashboard when you:<ul><li>Create an app in the UI and specify a value for **Application logo**.</li></ul>               | URL                                                                                            | TRUE       | FALSE    | FALSE      |
```
danielzeller-okta, line 1140: It feels a little odd to be referencing creating an app via the UI here. It also is making it sound a bit like if you created the app via the UI with a logo specified and then change `logo_uri` the new value would be used for the application dashboard, which is _not_ the case.
I think I would structure this by keeping `URL string that references a logo for the client. This value is used with the client consent dialog box during the client consent flow.` here in this table, and breaking out the rules/way it is populated when the client is created in the UI as a note below the table.

PR #1105, opened 2020-04-09
OKTA-282403: Prose layout for guides.

mrunalpatel-okta:
## Description:
- **What's changed?** 
prose layout for guide
no

### Resolves:

* [OKTA-282403](https://oktainc.atlassian.net/browse/OKTA-282403)


swiftone:
I would first merge to staging (if no one else is using it you can delete staging and re-create it with your branch) so we can test the site before live.

PR #1115, opened 2020-04-13
Clarification for salt property of hashed password object

motikagan-okta:
## Description:
- **What's changed?** Trying to make the description of the salt property clearer.

### DIFF
--- a/packages/@okta/vuepress-site/docs/reference/api/users/index.md
+++ b/packages/@okta/vuepress-site/docs/reference/api/users/index.md
@@ -3985,7 +3985,7 @@ Specifies a hashed password to import into Okta. This allows an existing passwor
 | -------- | ----- | ---------- |
 | algorithm  | String   | The algorithm used to generate the hash using the password (and salt, when applicable). Must be set to BCRYPT, SHA-512, SHA-256, SHA-1 or MD5. |
 | value      | String   | For SHA-512, SHA-256, SHA-1, MD5, This is the actual base64-encoded hash of the password (and salt, if used). This is the Base64 encoded `value` of the SHA-512/SHA-256/SHA-1/MD5 digest that was computed by either pre-fixing or post-fixing the `salt` to the `password`, depending on the `saltOrder`. If a `salt` was not used in the `source` system, then this should just be the the Base64 encoded `value` of the password's SHA-512/SHA-256/SHA-1/MD5 digest. For BCRYPT, This is the actual radix64-encoded hashed password. |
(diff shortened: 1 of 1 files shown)


motikagan-okta:
Screen Shot 2020-04-13 at 6 51 13 PM

packages/@okta/vuepress-site/docs/reference/api/users/index.md
```diff
@@ -3985,7 +3985,7 @@ Specifies a hashed password to import into Okta. This allows an existing passwor
...
 | value      | String   | For SHA-512, SHA-256, SHA-1, MD5, This is the actual base64-encoded hash of the password (and salt, if used). This is the Base64 encoded `value` of the SHA-512/SHA-256/SHA-1/MD5 digest that was computed by either pre-fixing or post-fixing the `salt` to the `password`, depending on the `saltOrder`. If a `salt` was not used in the `source` system, then this should just be the the Base64 encoded `value` of the password's SHA-512/SHA-256/SHA-1/MD5 digest. For BCRYPT, This is the actual radix64-encoded hashed password. |
-| salt       | String   | For salted algorithms, specifies the base64-encoded password salt used to generate the hash. For BCRYPT, specifies the radix64-encoded password salt used to generate the hash. Only required for salted algorithms. Length must be 22. |
+| salt       | String   | Only required for salted algorithms. For BCRYPT, specifies the radix64-encoded password salt used to generate the hash, which must be 22 characters long. For other salted algorithms, specifies the base64-encoded password salt used to generate the hash. |
```
rameshbhojan-okta, line 3988: nits : 
`password salt` -> `salt` (the word password with salt may be confusing).
`salted algorithms` -> `salted hashes` 
Overall, maybe rephrase the whole desc as follows : 

```
Only required for salted hashes. For BCRYPT, this specifies the radix64-encoded salt used to generate the hash, which must be 22 characters long. 
For other salted hashes (e.g SSHA-512/SSHA-256/SSHA-1/SMD5) , this specifies the base64-encoded salt used to generate the hash.
```

motikagan-okta: Made the edits, @rameshbhojan-okta, except for listing the the names of the example salted hashes. I'm worried it might be confusing unless we make it a complete list that we maintain.

rameshbhojan-okta: Sure @motikagan-okta , I am fine with that as that is what we actually changed from last time to move on to not list the hashes. :-) But, if you look at the suggestion above, it says `e.g.` so it doesn't have to be the complete list but is just a few examples, which I think is nice to have. Also, if you notice, all the examples there are prefixed with `S` say SMD5, which means salted MD5, so thought it might just be a few listed examples, because when we just say `For other salted hashes` it indirectly means all other salted hashes which again may not hold true, as we add more algorithms. So, I thought it would be nice to strike some balance. :-)

PR #1145, opened 2020-04-30
OKTA-291950: add Windows troubleshooting to README

bryanapellanes-okta:
## Description:
add Windows troubleshooting to README


README.md
```diff
@@ -62,6 +62,73 @@ This starts a preview server on your machine, and watches all files for changes.
 
 The preview server supports hot reloading. Once the server is running on your machine, any changes you make to Markdown content will appear automatically in your browser within a few seconds. Note that changes to page frontmatter or site configuration require you to stop and start the preview server.
 
+## Windows
```
laura-rodriguez, line 65: I suggest adding a `Troubleshooting` section, and put this content there. If anything else shows up (regardless of the env) in the future can be added into that section.

bryanapellanes-okta: Retty had the same Idea, I'll update.

README.md
```diff
@@ -62,6 +62,73 @@ This starts a preview server on your machine, and watches all files for changes.
...
 The preview server supports hot reloading. Once the server is running on your machine, any changes you make to Markdown content will appear automatically in your browser within a few seconds. Note that changes to page frontmatter or site configuration require you to stop and start the preview server.
 
+## Windows 
+
+If you are on a Windows machine you may experience the following issues.
+
+### EOL Error
+Because Windows handles line endings differently than *nix operating systems, you may see the following error after running `yarn dev`:
+
+```
```
laura-rodriguez, line 72: You can add bash/shell markdown here.

bryanapellanes-okta: I considered that; I wasn't sure if that was appropriate since the text is just output and not bash commands.

README.md
```diff
@@ -62,6 +62,73 @@ This starts a preview server on your machine, and watches all files for changes.
...
+    at async dev (C:\src\repos\okta-developer-docs\node_modules\@vuepress\core\lib\index.js:14:3)
+error Command failed with exit code 1.
+
+...
+```
+
+To resolve this issue run the following commands from a bash prompt (included with git for windows installation):
+
+```
+cd ./packages/\@okta/vuepress-theme-prose/
```
laura-rodriguez, line 126: You could use `yarn cwd` here to avoid having to change your current working directory. See https://stackoverflow.com/questions/46891622/run-yarn-in-a-different-path.

README.md
laura-rodriguez: I think it's OK since this is the output you see in your console after running `yarn dev`.

README.md
bryanapellanes-okta: I'll test that locally and update.

PR #1170, opened 2020-05-08
Added groups as reserved access token claim for Token Inline Hook

dragos-okta:
## Description:
- **What's changed?** Added groups as reserved access token claim for Token Inline Hook


dragos-okta:
This can be confirmed through the following payload

`{
    "commands": [{
        "type": "com.okta.access.patch",
        "value": [{
            "op": "replace",
            "path": "/claims/groups",
            "value": ["a", "b", "c", "d"]
        }]
    }]
}`

PR #1198, opened 2020-05-20
Merging doc update bash scripts

jakubvul-okta:
## Description:
- **What's changed?** Three bash scripts have been consolidated into one by someone who had no idea what they were doing.

### Resolves:

* N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     "posttest": "yarn stop",
     "stop": "if-env CI=true && exit 0 || yarn test:stop-web-driver || echo 'done'",
     "stop-all": "if-env CI=true && exit 0 || pkill -f 'yarn' || echo 'done'",
+    "update-docs": "yarn update-event-types && yarn update-error-codes && yarn update-oauth2-error-codes",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -24,9 +24,7 @@
     "posttest": "yarn stop",
     "stop": "if-env CI=true && exit 0 || yarn test:stop-web-driver || echo 'done'",
     "stop-all": "if-env CI=true && exit 0 || pkill -f 'yarn' || echo 'done'",
-    "update-event-types": "bash scripts/update-event-types.sh",
-    "update-error-codes": "bash scripts/update-error-codes.sh",
-    "update-oauth2-error-codes": "bash scripts/update-oauth2-error-codes.sh"
+    "update-docs": "bash scripts/update-docs.sh"
```
bretterer, line 27: Instead of making your own script, why not just do `"update-docs": "yarn update-event-types && yarn update-error-codes && yarn update-oauth2-error-codes"` that way we can update each one individually still if we want :)

PR #1231, opened 2020-06-08
Setting up Quickstart Redirects to Guides

bretterer:
### Resolves:

* [OKTA-248446](https://oktainc.atlassian.net/browse/OKTA-248446)


packages/@okta/vuepress-theme-prose/components/Quickstart.vue
```diff
@@ -1,265 +1,87 @@
...
         activeServer: null,
-        activeServerComponent: null,
         activeFramework: null,
-        otpItems: [
-          {'title': 'Create an Account', 'path': '#create-an-okta-account'},
-          {'title': 'Client Setup', 'path': '#client-setup'},
-          {'title': 'Server Setup', 'path': '#server-setup'}
-        ]
+        mappings: {
+          '#/okta-sign-in-page/nodejs/express' : '/docs/guides/sign-into-web-app/nodeexpress/before-you-begin/',
```
swiftone, line 11: Not suggesting this, just putting it out there - if this page just booted you to a url that had all the elements from the hash but didn't use a hash (e.g. /quickstarts/#/okta-sign-in-page/nodejs/express => /quickstarts/okta-sign-in-page/nodejs/express) then we could have the actual redirects in conductor.yml so they'd have the same management.

Downsides: 
- two redirects instead of one
- conductor.yml may be consolidated and consistent, but that doesn't make it desirable
- you might have to create any intermediate directories for the conductor-file-based approach to work (you could resolve this by not having more paths in the original redirect). 

I'm happy with this approach, but I wanted to make sure the alternative was considered.

tests/selenium/conf.js
```diff
@@ -6,7 +6,7 @@ const config = {
     retries: 4
   },
   specs: ['spec/*.js'],
-  exclude: ['spec/quickstarts-spec.js'],
+  // exclude: ['spec/quickstarts-spec.js'],
```
robertjd, line 9: Nit, remove

packages/@okta/vuepress-theme-prose/components/Quickstart.vue
```diff
@@ -1,265 +1,87 @@
...
+          '#/react-native/java/generic' : '/docs/guides/sign-into-mobile-app/react-native/before-you-begin/',
+          '#/react-native/php/generic' : '/docs/guides/sign-into-mobile-app/react-native/before-you-begin/',
+          '#/react-native/dotnet/aspnetcore' : '/docs/guides/sign-into-mobile-app/react-native/before-you-begin/',
+          '#/react-native/dotnet/aspnet4' : '/docs/guides/sign-into-mobile-app/react-native/before-you-begin/',
+        }
       }
     },
 
     beforeMount: function () {
       if(window.location.hash == "") {
```
swiftone, line 72: I don't see anything alarming, but just double-checking: have you tested this with a build and static server to ensure server-side rendering doesn't break anything here?

tests/selenium/conf.js
swiftone, line 9: just remove now?

packages/@okta/vuepress-theme-prose/components/Quickstart.vue
swiftone, line 72: Your python heart may quail, but in cases like this I generally recommend using truthy/falsy
```
if( !window.location.hash ) { 
```

As the `==` draws the eye away from the active part that matters when skimming code

packages/@okta/vuepress-theme-prose/components/Quickstart.vue
```diff
@@ -1,265 +1,87 @@
...
-          if(server.name == this.activeServer) {
-            return true
-          }
-        })
-
-        if (!found) {
-          this.activeServer = this.defaultServer.name
-        }
+      let redirecturl = this.mappings[window.location.hash]
+      if(redirecturl == undefined) {
```
swiftone, line 78: `if( !redirecturl ) { `

Also: Why not `redirectUrl` (capital U)?

swiftone: Actually, this could all be:
```
const redirectUrl = this.mappings[window.location.hash] || '/docs/guides';
window.location.replace(redirectUrl);
```

tests/selenium/conf.js
bretterer: Ugh... I had removed it, but didn't hit save

bretterer: Ugh... I had removed it, but didn't hit save

PR #1235, opened 2020-06-09
Read-Only Mode and the Password Import Inline Hook

motikagan-okta:
## Description:
- **What's changed?** Updating Password Import Inline Hook documentation to mention the possibility of passwords not being saved if an org is in read-only mode, and to provide information on using an Event Hook to verify status before deleting passwords from an existing user store.

### Resolves:

* [OKTA-283131](https://oktainc.atlassian.net/browse/OKTA-283131)


packages/@okta/vuepress-site/docs/reference/password-hook/index.md
```diff
@@ -161,5 +161,5 @@ Normally, if your external service responds to Okta indicating that the credenti
 
 Because of the posssiblity of read-only mode, permanent deletion of user passwords from your existing user store should not be performed until success of the password migration can be verified. An Okta System Log [Event](/docs/reference/api/event-types/), `user.import.password`, is available for this purpose. An event of this type is created every time a Password Import Inline Hook is fired, with its `Event.Outcome` property providng a status of `FAILURE` or `SUCCESS` for the password import operation. If the status is `SUCCESS`, Okta has successfully saved the end user's password, and it's safe to delete it from your previous user store.
```
amoghvasekar-okta, line 162: Is this technically accurate @kenwang-okta @rameshbhojan-okta ? I was under the assumption that password import hooks work in Read-Only mode?
I agree we should still have a note around not deleting password until event is fired.

packages/@okta/vuepress-site/docs/reference/password-hook/index.md
```diff
@@ -151,5 +151,15 @@ To enable a Password Import Inline Hook, you first need to register your externa
...
+
+## Removing Password from Existing User Store
+
+Because of the posssiblity of read-only mode, permanent deletion of user passwords from your existing user store should not be performed until success of the password migration can be verified. An Okta System Log [Event](/docs/reference/api/event-types/), `user.import.password`, is available for this purpose. An event of this type is created every time a Password Import Inline Hook is fired, with its `Event.Outcome` property providng a status of `FAILURE` or `SUCCESS` for the password import operation. If the status is `SUCCESS`, Okta has successfully saved the end user's password, and it's safe to delete it from your previous user store.
```
jakubvul-okta, line 162:
```suggestion
Because of the possibility of read-only mode, permanent deletion of user passwords from your existing user store should not be performed until success of the password migration can be verified. An Okta System Log [Event](/docs/reference/api/event-types/), `user.import.password`, is available for this purpose. An event of this type is created every time a Password Import Inline Hook is fired, with its `Event.Outcome` property providing a status of `FAILURE` or `SUCCESS` for the password import operation. If the status is `SUCCESS`, Okta has successfully saved the end user's password, and it's safe to delete it from your previous user store.
```

packages/@okta/vuepress-site/docs/reference/password-hook/index.md
```diff
@@ -151,5 +151,15 @@ To enable a Password Import Inline Hook, you first need to register your externa
...
+When the end user that you have added attempts to sign in to Okta for the first time, the hook is triggered and Okta calls your external service, sending it the credentials that end user provided. Your service can check the credentials and respond with a command to indicate to Okta whether the credentials are valid or not.
+
+## Password Inline Hook and Okta Service Mode
+
+Normally, if your external service responds to Okta indicating that the credentials are valid, Okta saves the password and can authenticate the user independently from then on. However, if your Okta org is in  *read-only mode* (see [Okta Service Mode](https://help.okta.com/en/prod/okta_help_CSH.htm#ext_ref_service_op_mode)) at the time the end user signs in, then saving the password isn't possible, abd the next time the end user attempts to sign in, the Password Import Inline Hook needs to be called again.
```
jakubvul-okta, line 158:
```suggestion
Normally, if your external service responds to Okta indicating that the credentials are valid, Okta saves the password and can authenticate the user independently from then on. However, if your Okta org is in  *read-only mode* (see [Okta Service Mode](https://help.okta.com/en/prod/okta_help_CSH.htm#ext_ref_service_op_mode)) at the time the end user signs in, then saving the password isn't possible, and the next time the end user attempts to sign in, the Password Import Inline Hook needs to be called again.
```

motikagan-okta: Fixed in a separate commit.

packages/@okta/vuepress-site/docs/reference/password-hook/index.md
motikagan-okta: Fixed in a separate commit.

packages/@okta/vuepress-site/docs/reference/password-hook/index.md
```diff
@@ -27,7 +27,7 @@ The Password Import Inline Hook enables migration of users from another data sto
 
 The Password Import Inline Hook is triggered when the end user tries to sign in to Okta for the first time. Okta sends your external service the password that the user supplied. Your external service then needs to send a response to Okta indicating whether the password supplied by the end user is valid or not.
 
-If the password is valid, Okta can authenticate the user independently from then on.
+If the password is valid, Okta can usually authenticate the user independently from then on, but see [Password Inline Hook and Okta Service Mode](#password-inline-hook-and-okta-service-mode) and [Removing Password from Existing User Store](#removing-password-from-existing-user-store) below.
```
wdawson, line 30: This phrasing feels awkward. I'd suggest something like this:
> When your service returns a response that indicates the password was valid, Okta will set the password for the user and will not call your service again. In cases where the Okta service is disrupted or degraded, Okta may not be able to set the password and will call your service on subsequent authentication attempts until the password is successfully set. See [links] for details.

packages/@okta/vuepress-site/docs/reference/password-hook/index.md
rameshbhojan-okta: Yes, technically we import it even in RO mode, and only don't do anything in the `SAFE` mode. So, maybe we shouldn't specify `read-only mode` there and instead just state it like generically something along the lines of `Depending on the mode of operation`

motikagan-okta: @rameshbhojan-okta, changed to:
"Because of the possibility of your org being in a special service mode, permanent deletion of user passwords from your existing user store should not be performed until success of the password migration can be verified."

packages/@okta/vuepress-site/docs/reference/password-hook/index.md
```diff
@@ -27,7 +27,7 @@ The Password Import Inline Hook enables migration of users from another data sto
 
 The Password Import Inline Hook is triggered when the end user tries to sign in to Okta for the first time. Okta sends your external service the password that the user supplied. Your external service then needs to send a response to Okta indicating whether the password supplied by the end user is valid or not.
 
-If the password is valid, Okta can authenticate the user independently from then on.
+When your service returns a response that indicates that the password is valid, Okta sets the password for the user and won't call your service again. However, if Okta service is temporarily disrupted or degraded, Okta might not be able to set the password, and will need to call your service on subsequent authentication attempts until the password is successfully set. See [Password Inline Hook and Okta Service Mode](#password-inline-hook-and-okta-service-mode) and [Removing Password from Existing User Store](#removing-password-from-existing-user-store) for details.
```
rameshbhojan-okta, line 30: nit : not sure if we should use the word `set` password or `import` password. Same way, in the end - until the password is successfully `set` vs. until the password is successfully `imported`.

packages/@okta/vuepress-site/docs/reference/password-hook/index.md
```diff
@@ -151,6 +151,16 @@ To enable a Password Import Inline Hook, you first need to register your externa
...
+
+## Removing Password from Existing User Store
+
+Because of the possibility of your org being in a special service mode, permanent deletion of user passwords from your existing user store should not be performed until success of the password migration can be verified. An Okta System Log [Event](/docs/reference/api/event-types/), `user.import.password`, is available for this purpose. An event of this type is created every time a Password Import Inline Hook is fired, with its `Event.Outcome` property providing a status of `FAILURE` or `SUCCESS` for the password import operation. If the status is `SUCCESS`, Okta has successfully saved the end user's password, and it's safe to delete it from your previous user store.
```
rameshbhojan-okta, line 162: nit : `migration` -> `import`

PR #1258, opened 2020-06-23
Adding Password Hook to Sidebar Links

motikagan-okta:
## Description:
- **What's changed?** Adding a link to the reference page about the Password Import Inline Hook to the left-side nav sidebar.


packages/@okta/vuepress-site/.vuepress/nav/reference.js
```diff
@@ -37,6 +37,7 @@ module.exports = [
     { title: 'Registration Hook', path: '/docs/reference/registration-hook/'},
     { title: 'Token Hook', path: '/docs/reference/token-hook/'},
     { title: 'Okta Expression Language', path: '/docs/reference/okta-expression-language/'},
+    { title: 'Password Hook', path: '/docs/reference/password-hook/'},
```
IanHakes-okta, line 40: Is there a reason why Password Hook comes after the OEL, instead of with the other "X Hook" reference items?

motikagan-okta: I assume it should be in alphabetical order, so, have now hanged those other hook links to be in order.

PR #1280, opened 2020-06-30
OKTA-307115 move publish to npm-topic

timgaskin-okta:
## Description:
- **What's changed?** 
Moves the repackaging and publishing flows to the new standard, repackaging using ci-append-sha and publishing to npm-topic. This change will require the artifacts to be promoted in bacon before they can be released.

### Resolves:

* [OKTA-307115](https://oktainc.atlassian.net/browse/OKTA-307115)


uguraktepe-okta:
conductor is configured to deploy from `npm-okta`. 
afaik, `npm-topic` is not configured to replicate to production artifactory servers.
this will most likely break the deployments.

jamesbui-okta:
Once `npm-release` is replicated to production artifactory servers, this repo will have to go through promotions in order to get their topic artifacts from `npm-topic` to `npm-release`. At which point the deployments will behave as normal.

bretterer:
@jamesbui-okta so will it all still be automated where we don't have to interact with bacon at all like we do today?

jamesbui-okta:
> @jamesbui-okta so will it all still be automated where we don't have to interact with bacon at all like we do today?

I just pushed a change so promotion is automatic.

PR #1517, opened 2020-10-02
Fixed active states for nav bar

AndriiDovhopolov-okta:
## Description:
- Active states for navbar were fixed. After discussing with @alexdahl-okta  there was decided that each parent of active link will have bold text and active item will remain bold to: 
image

- Fixed issue with relative links and exact links (active link style was applied to parent part of a link, even if link was not currently active - this why first item of sidebar was always bold, see "Languages & SDKs" and "Add user auth to your angular app"). Screenshot on before state, current state on previous screenshot: 
image

- Fixed "keys" properties for Vue for render directive - now link names are used for "keys" instead of array indexes (when we have non-flat arrays we can not use indexes - several elements will have same "key" properties, this can cause side effects when components are updated) 

- Fixed issue with double active links in Concepts section. 

Before: 

bugged-gif

After:

fixed-gif

- Fixed issue with implementing same set of styles twice. 
Before: 
image
After:
image

### Resolves:

* [OKTA-320822](https://oktainc.atlassian.net/browse/OKTA-320822)

### DIFF
--- a/packages/@okta/vuepress-theme-prose/components/Sidebar.vue
+++ b/packages/@okta/vuepress-theme-prose/components/Sidebar.vue
@@ -1,14 +1,14 @@
 <template>
   <aside class="landing-navigation" :class="{active: sidebarActive}">
     <ul class="landing">
-      <SidebarItem v-for="(link, key) in navigation" :key="key" :link="link" />
+      <SidebarItem v-for="link in navigation" :key="link.title" :link="link" />
     </ul>
   </aside>
 </template>
 
 <script>
 import _ from "lodash";
-import { getGuidesInfo, guideFromPath } from "../util/guides";
+import { getGuidesInfo, guideFromPath } from "../util/guides";  
 
 export default {
   name: "Sidebar",
--- a/packages/@okta/vuepress-theme-prose/components/SidebarItem.vue
+++ b/packages/@okta/vuepress-theme-prose/components/SidebarItem.vue
@@ -10,10 +10,10 @@
             </svg>
 
             <div v-if="link.path">
-                <router-link :to="link.path" @click="setData" :class="{'router-link-active': link.imActive, 'router-link-exact-active': link.imActive}">{{link.title}}</router-link>
+                <router-link :to="link.path" @click="setData" :class="{'router-link-active': link.imActive, 'router-link-exact-active': iHaveChildrenActive}">{{link.title}}</router-link>
             </div>
             <div v-else>
-                <div class="is-link" @click="toggle">{{link.title}}</div>
+                <div :class="{'is-link':true, 'router-link-active': iHaveChildrenActive}" @click="toggle">{{link.title}}</div>
             </div>
         </div>
 
--- a/packages/@okta/vuepress-theme-prose/components/SidebarItem.vue
+++ b/packages/@okta/vuepress-theme-prose/components/SidebarItem.vue
@@ -9,16 +9,17 @@
                 <path d="M31.3 192h257.3c17.8 0 26.7 21.5 14.1 34.1L174.1 354.8c-7.8 7.8-20.5 7.8-28.3 0L17.2 226.1C4.6 213.5 13.5 192 31.3 192z"/>
             </svg>
 
-            <div v-if="link.path">
-                <router-link :to="link.path" @click="setData" :class="{'tree-nav-link': true, 'router-link-active': link.imActive, 'router-link-exact-active': link.imActive}">{{link.title}}</router-link>
+             <div v-if="link.path">
+                <router-link :to="link.path" exact @click="setData" class="tree-nav-link">{{link.title}}</router-link>
             </div>
+
             <div v-else>
                 <div :class="{'is-link':true, 'router-link-active': iHaveChildrenActive}" @click="toggle">{{link.title}}</div>
             </div>
         </div>
 
         <ul v-if="link.subLinks" class="sections" v-show="iHaveChildrenActive">
-            <SidebarItem v-for="(sublink, key) in link.subLinks" :key="key" :link="sublink" />
+            <SidebarItem v-for="sublink in link.subLinks" :key="sublink.title" :link="sublink" />
         </ul>
       </li>
 </template>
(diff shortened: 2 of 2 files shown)


packages/@okta/vuepress-theme-prose/components/Sidebar.vue
```diff
@@ -1,14 +1,14 @@
 <template>
   <aside class="landing-navigation" :class="{active: sidebarActive}">
     <ul class="landing">
-      <SidebarItem v-for="(link, key) in navigation" :key="key" :link="link" />
+      <SidebarItem v-for="link in navigation" :key="link.title" :link="link" />
```
bretterer, line 4: What was the purpose of changing the `:key` item here?

packages/@okta/vuepress-theme-prose/components/SidebarItem.vue
```diff
@@ -9,16 +9,17 @@
                 <path d="M31.3 192h257.3c17.8 0 26.7 21.5 14.1 34.1L174.1 354.8c-7.8 7.8-20.5 7.8-28.3 0L17.2 226.1C4.6 213.5 13.5 192 31.3 192z"/>
             </svg>
 
-            <div v-if="link.path">
-                <router-link :to="link.path" @click="setData" :class="{'router-link-active': link.imActive, 'router-link-exact-active': link.imActive}">{{link.title}}</router-link>
+             <div v-if="link.path">
+                <router-link :to="link.path" exact @click="setData" class="tree-nav-link">{{link.title}}</router-link>
```
bretterer, line 13: What does `exact` do?

packages/@okta/vuepress-theme-prose/components/SidebarItem.vue
```diff
@@ -9,16 +9,17 @@
...
+
             <div v-else>
-                <div class="is-link" @click="toggle">{{link.title}}</div>
+                <div :class="{'is-link':true, 'router-link-active': iHaveChildrenActive}" @click="toggle">{{link.title}}</div>
             </div>
         </div>
 
         <ul v-if="link.subLinks" class="sections" v-show="iHaveChildrenActive">
-            <SidebarItem v-for="(sublink, key) in link.subLinks" :key="key" :link="sublink" />
+            <SidebarItem v-for="sublink in link.subLinks" :key="sublink.title" :link="sublink" />
```
bretterer, line 22: Same questions

bretterer (changes requested):
Left some comments before approval

packages/@okta/vuepress-theme-prose/components/Sidebar.vue
AndriiDovhopolov-okta: @bretterer mentioned that in PR description:
>when we have non-flat arrays using indexes is not a best solution - several elements will have same "key" properties, this can cause side effects when components are updated and re-rendered. 

In detail - Vue optimizes DOM operations using key properties(basically, as any other framework, such as Angular or React). When Vue see several elements with same keys it will not be happy about that, will get confused and will swap whole DOM representation, instead of particular element with particular key ID. 

There is a link to docs, it's not super in-depth, but explains better than me
https://vuejs.org/v2/guide/list.html#Maintaining-State

packages/@okta/vuepress-theme-prose/components/SidebarItem.vue
AndriiDovhopolov-okta: Here is link for Vue documentation  - https://router.vuejs.org/api/#exact

In general, `exact` will apply `router-link-active` class (default class name provided by Vue) ONLY if the whole current URL will be exactly equal to `:to` directive of `<router-link>`

E.g. we have such URL structure: 

- /parent/
- /parent/first-child/
- /parent/second-child/
- .....
- /parent/nth-child/

`<router-link>` provides `router-link-active` class and URL tracking feature from the box by default. If we will not use `exact` directive, `<router-link>` will treat /parent link as active link, even if current active link is /parent/nth-child, because in provided structure all routes from /first-child/ to /nth-child/ will be treated as siblings and /parent/ will be treated as parent.  

This why there was `router-link-active` class added to each first item in each sidebar sections (Concepts, Languages and SDKs, References). 

Also, due to out-of-the box default `<router-link>` features there is no need to manually add or remove classes to/from this component in Vue, see issue with dual class sets in description of PR.

alexdahl-okta:
Looks great to me! 🎉

packages/@okta/vuepress-theme-prose/components/SidebarItem.vue
AndriiDovhopolov-okta: Look up to previous answer.

packages/@okta/vuepress-theme-prose/components/SidebarItem.vue
```diff
@@ -9,16 +9,17 @@
...
 
-            <div v-if="link.path">
-                <router-link :to="link.path" @click="setData" :class="{'router-link-active': link.imActive, 'router-link-exact-active': link.imActive}">{{link.title}}</router-link>
+             <div v-if="link.path">
+                <router-link :to="link.path" exact @click="setData" class="tree-nav-link">{{link.title}}</router-link>
             </div>
+
             <div v-else>
-                <div class="is-link" @click="toggle">{{link.title}}</div>
+                <div :class="{'is-link':true, 'router-link-active': iHaveChildrenActive}" @click="toggle">{{link.title}}</div>
```
oleksandrpravosudko-okta, line 17: nit: since we have figured `router-link-active` is a reserved class name in `router-link`, let's pick a different name (for example, `link-ancestor-active`)

oleksandrpravosudko-okta:
this might be a good candidate for adding a selenium test (ancestors and selected item are highlighted with bold font)

bretterer (approved):
LGTM!  Thanks for the explanations!

bretterer:
After merging #1528, I now have a conflict.  can you fix @oleksandrpravosudko-okta?

oleksandrpravosudko-okta:
@bretterer - https://github.com/okta/okta-developer-docs/pull/1531 contains a copy of  this change set (I don't have write access for Andriy's fork)

PR #1526, opened 2020-10-04
minor fix to those inconsistent code block

haishengwu-okta:
## Description:

- **What's changed?**
    1. Change code block language to lower case (although it doesn't matter in terms of functionality, turns out the conversion is to use lower case
    2. Fix a few some blocks which language apparently are not correct. (e.g. shall be bash instead of json, shall be json instead of XML, etc)
    3. Fix a few JSON block that JSON value is malformatted.
    4. whitespace clean up.

### Resolves:


susanharper-okta (approved):
LGTM - thank you for fixing all of these!

PR #1554, opened 2020-10-12
Add conditional netlify build

robertjd:
Deploys the site to netlify, if the branch is in origin and the NFDEPLOY text is present in the commit message

Depends on the following env vars which have been added to travis via the UI:

NETLIFY_AUTH_TOKEN
NETLIFY_SITE_ID


bretterer (approved):
This looks good to me.

Thanks!

PR #1563, opened 2020-10-13
Update Angular and React Auth JS Guides

mraible:
## Description:
- **What's changed?** Updated Angular guide after QA. Attempted to update React guide, but failed. This PR is to track my progress until it's fixed.

### Resolves:

* [OKTA-337095](https://oktainc.atlassian.net/browse/OKTA-337095)


mraible:
After creating all the code for the React + Auth JS (functional) guide, when I click on **Login**, I just get a `Loading...` page.

Screen Shot 2020-10-13 at 2 41 42 PM

If I refresh my browser, it renders the form, but when I enter my credentials, the following error results:

Screen Shot 2020-10-13 at 2 41 05 PM

packages/@okta/vuepress-site/code/react/okta_react/index.md
```diff
@@ -346,14 +351,12 @@ export default withOktaAuth(class SignIn extends Component {
 The component for this route (LoginCallback) comes with `@okta/okta-react`. It handles token parsing, token storage, and redirecting to a protected page if one triggered the sign in.
 
 ### Connect the Routes
-Update `src/App.jsx` to include your project components and routes. `Security` is the component that controls the authentication flows, so it requires your OpenId Connect configuration. By default, `@okta/okta-react` redirects to Okta's sign in page when the user isn't authenticated.
+Update `src/App.js` to include your project components and routes. `Security` is the component that controls the authentication flows, so it requires your OpenId Connect configuration. By default, `@okta/okta-react` redirects to Okta's sign in page when the user isn't authenticated.
```
swiftone, line 354: I accept this change, but I feel switching from .js to .jsx is one of two notable mis-steps Facebook has made with React progress.  It's a pure removal of information.

`</soapbox>`

packages/@okta/vuepress-site/code/react/okta_react/index.md
```diff
@@ -346,14 +351,12 @@ export default withOktaAuth(class SignIn extends Component {
...
 
 In this example, `onAuthRequired` is overridden to redirect to the custom sign-in route instead, which requires a component that is a descendent of Router to have access to `react-router`'s `history`.  Other router libraries will have their own methods of managing browser history:
 
-`src/App.jsx` and `src/AppWithRouterAccess.jsx` using function-based components:
+Update `src/App.js` to use function-based components:
```
swiftone, line 358: create-react-app is already using function-based components.  Here the guide is electing to having the dev-created components be function-based, so I don't really like "Update...to use"

packages/@okta/vuepress-site/code/react/okta_react/index.md
```diff
@@ -385,25 +389,24 @@ const AppWithRouterAccess = () => {
   };
 
   return (
-    <Security issuer='https://${yourOktaDomain}/oauth2/default'
```
swiftone, line 388: My embarrassment is considerable.

packages/@okta/vuepress-site/code/react/okta_react/index.md
```diff
@@ -385,25 +389,24 @@ const AppWithRouterAccess = () => {
...
+    <Security issuer='https://{yourOktaDomain}/oauth2/default'
```
aarongranick-okta, line 392: I think the routes should be wrapped in a `<Switch>` component. Otherwise `/login` may render on `/login/callback`. Probably /login/callback needs to come before /login for the same reason

packages/@okta/vuepress-site/code/react/okta_react/index.md
```diff
@@ -442,14 +443,14 @@ export default withRouter(class AppWithRouterAccess extends Component {
 
   render() {
     return (
-      <Security issuer='https://${yourOktaDomain}/oauth2/default'
+      <Security issuer='https://{yourOktaDomain}/oauth2/default'
```
aarongranick-okta, line 446: Same comment here about switch and multiple components rendering on a single route

packages/@okta/vuepress-site/code/react/okta_react/index.md
mraible: The reason I changed it is because `.js` files are created by default. It's either "update App.js" or "delete App.js and create App.jsx" or "rename App.js to App.jsx". Using "update" was the easiest, so that's what I used. I'm not tied to it.

packages/@okta/vuepress-site/code/react/okta_react/index.md
mraible: I tried moving `/login/callback` first and it doesn't help. Do you have an example of using `<Switch>`? Maybe I should just use `/callback`? :trollface:

packages/@okta/vuepress-site/code/react/okta_react/index.md
swiftone: No, since CRA gives us .js, this is the better option.  I just feel regret at the necessary acceptance.

aarongranick-okta:
@mraible The cryptic error about "default.a" is probably because you are using `@okta/okta-auth-js` version 4. So it should be `import { OktaAuth } from '@okta/okta-auth-js'`

mraible:
> So it should be import { OktaAuth } from '@okta/okta-auth-js'

Yep, that worked. 👍

swiftone:
Could I ask you to add a "This guide is written for okta-auth-js version 4.x and okta-react(angular) version NNN.x" (where .x is actually ".x")?  The switch to the okta-auth-js 4.x is particularly important, as you've discovered, so this is a good place to start calling our versions out.

packages/@okta/vuepress-site/code/react/okta_react/index.md
swiftone: ...and I wanted someone to hear my reluctance.  That was apparently important to me.

aarongranick-okta:
@mraible https://reactrouter.com/core/api/Switch

mraible:
I tried this:

```
<Security issuer='https://dev-162929.okta.com/oauth2/default'
          clientId='0oa96n9rA7zUhy1Uu5d5'
          redirectUri={window.location.origin + '/login/callback'}
          onAuthRequired={onAuthRequired}
          pkce={true} >
  <Switch>
    <Route path='/' exact={true} component={Home} />
    <SecureRoute path='/protected' component={Protected} />
    <Route path='/login/callback' component={LoginCallback} />
    <Route path='/login' render={() => <SignIn issuer='https://dev-162929.okta.com/oauth2/default' />} />
  </Switch>
</Security>
```

It doesn't seem to help.

mraible:
> Could I ask you to add a "This guide is written for okta-auth-js version 4.x and okta-react(angular) version NNN.x" (where .x is actually ".x")?

Like this?

Screen Shot 2020-10-13 at 3 34 25 PM

mraible:
@jakubvul-okta Do we have a style for calling out version numbers? I just used a blockquote.

jakubvul-okta:
> @jakubvul-okta Do we have a style for calling out version numbers? I just used a blockquote.

We don't, and that seems reasonable.

mraible:
After thinking about this a bit more, I think we should be explicit. We should have a callout at the top and specify the version for `npm install` commands. 

The reason: things break in patch releases. It happens. 

If we use specific version numbers, we'll know it works because someone QA'd it. 

Unless we can automate creating the code in the guides as part of the SDK releases, they're always going to be out of sync. That's why I think our guides should reflect "last tested with". Adding a callout at the top and specific version numbers in the commands would do this, IMO.

packages/@okta/vuepress-site/code/react/okta_react/index.md
mraible: Should I keep the same words as before? The only thing I don't like about it is there are two code blocks with no words in between them.

mraible:
I went through the Sign-In Widget guide for React this morning and updated it. I ran into the same issue that I did with Auth JS. When you first click on Login, you get a blank screen. If you refresh the page, it works.

I created GitHub repos with both examples:

* https://github.com/mraible/okta-react-auth-js
* https://github.com/mraible/okta-react-signin-widget

mraible:
I was able to fix the Login button issue with @shuowu-okta's help.

- For Auth JS: https://github.com/mraible/okta-react-auth-js/commit/3e9818e25acf67c8019896b8692c5cb9bc3f040c
- For Sign-In Widget: https://github.com/mraible/okta-react-signin-widget/commit/ba27df55fb0da4887f0eb2bdf90c9154cd2cc87a

I also created Auth JS repos from QA'ing Angular and Vue:

- Angular: https://github.com/mraible/okta-angular-auth-js
- Vue: https://github.com/mraible/okta-vue-auth-js

mraible:
There is a console warning from React when you go to `/login` in the Sign-In Widget example.

image

packages/@okta/vuepress-site/code/react/okta_react_sign-in_widget/index.md
```diff
@@ -15,6 +15,8 @@ This guide will walk you through integrating authentication into a React app wit
 6. [Connect the Routes](#connect-the-routes)
 7. [Start Your App](#start-your-app)
 
+> This guide was last tested with @okta/okta-signin-widget v4.5.1 and @okta/okta-react v3.0.8.
```
susanharper-okta, line 18: I think this format is more clear and doesn't leave the user wondering well - is there a newer version that needs to be tested with this guide now? Should I use these steps, then? I think the below format leaves less to question:

> This guide is for @okta/okta-signin-widget v4.5.1 and @okta/okta-react version 3.0.8.

packages/@okta/vuepress-site/code/react/okta_react_sign-in_widget/index.md
```diff
@@ -46,28 +48,25 @@ If you need more information, see [the Create React App getting started guide](h
...
-A simple way to add authentication into a React app is using the [Okta Sign-In Widget](/code/javascript/okta_sign-in_widget/) library. We can install it via `npm`:
+To provide a fully-featured and customizable sign-in experience, the [Okta Sign-In Widget](/code/javascript/okta_sign-in_widget/) is available to handle User Lifecycle operations, MFA, and more. You can install it using `npm`:
 
 ```bash
-cd okta-app
 npm install @okta/okta-signin-widget
 ```
 
-We'll also need `@okta/okta-react` and `react-router-dom` to manage our routes:
+You'll also need `@okta/okta-react` and `react-router-dom` to manage our routes:
```
susanharper-okta, line 57: future to present:

You also need....

packages/@okta/vuepress-site/code/react/okta_react_sign-in_widget/index.md
```diff
@@ -215,21 +207,17 @@ export default withOktaAuth(class Login extends Component {
...
-
-### `/implicit/callback`
+### `/login/callback`
 
 The component for this route (LoginCallback) comes with `@okta/okta-react`. It handles token parsing, token storage, and redirecting to a protected page if one triggered the login.
 
 ### Connect the Routes
 
-Our example is using `react-router-dom`. By default you can include your components and Routes in `src/App.js`. If you need access to particular router properties, such as the `history` object that we use to override the default login flow, you need to create a wrapper component around `<Router>`.
+Our example is using `react-router-dom`. By default you can include your components and Routes in `src/App.js`. If you need access to particular router properties, such as the `history` object that's used to override the default login flow, you need to create a wrapper component around `<Router>`.
```
susanharper-okta, line 216: ......default sign-in flow......

susanharper-okta (commented):
couple of style things

packages/@okta/vuepress-site/code/react/okta_react/index.md
mraible: Used `history.push('/login')` to solve.

mraible:
@susanharper-okta I updated this PR with your suggestions. Back to you!

packages/@okta/vuepress-site/code/react/okta_react_sign-in_widget/index.md
```diff
@@ -247,11 +235,11 @@ class App extends Component {
 export default App;
 ```
 
-Create `src/AppWithRouterAccess.js` and include your project components and routes. `Security` is the component that controls the authentication flows, so it requires your OpenId Connect configuration. By default, `@okta/okta-react` redirects to Okta's login page when the user isn't authenticated. In this example, `onAuthRequired` is overridden to redirect to the custom login route instead:
+Create `src/AppWithRouterAccess.js` and include your project components and routes. `Security` is the component that controls the authentication flows, so it requires your OpenID Connect configuration. By default, `@okta/okta-react` redirects to Okta's login page when the user isn't authenticated. In this example, `onAuthRequired` is overridden to redirect to the custom login route instead:
```
susanharper-okta, line 238: ....redirects to Okta's login page when.... ->

...redirects to Okta's sign-in page when....

....to the custom login route instead: ->
...to the custom sign-in route instead:

Sorry, I missed a couple

susanharper-okta (approved):
LGTM - I missed two things, otherwise good - so I'm approving to not hold you up.

packages/@okta/vuepress-site/code/react/okta_react_sign-in_widget/index.md
mraible: I fixed these. Can you please merge when tests pass? I don't have permission.

PR #1581, opened 2020-10-21
Expand Widget initialization example

mraible:
## Description:
- **What's changed?** Adds all HTML necessary for the Widget's initialization.


robertjd:
@mraible IIRC we're also trying to merge some sample code into the Widget repo to be the "canonical" source of truth for reference code?  Is this code similar enough to that, so that we could refactor down to one location when we get there? CC @aarongranick-okta @haishengwu-okta - basically I'm asking a meta question about this piece of content and where sample code is meant to be living, I've seen several conversations around this topic

robertjd:
Bumping this one up - it sounds like we're also about to make another widget release, so this will be out of date yet again.  Do we need to met to discuss widget samples strategy?

PR #1593, opened 2020-10-29
slight update to content

susanharper-okta:
## Description:
- **What's changed?** Provided URLs to how to content

### Resolves:

* [OKTA-341584](https://oktainc.atlassian.net/browse/OKTA-341584)


jakubvul-okta (approved):

jakubvul-okta:
Let's hold off until we get Padma's approval.

codepadma (commented):
@susanharper-okta For [Client Credentials Flow](https://developer.okta.com/docs/guides/implement-client-creds/use-flow/), could you change 'Okta Authorization Server's' to 'Custom Authorization Server's ' and change the link to point to [https://developer.okta.com/docs/concepts/auth-servers/#custom-authorization-server](https://developer.okta.com/docs/concepts/auth-servers/#custom-authorization-server) ? (as this flow cannot be used by end user, hence oidc doesn't come into play for this flow)

It might also be worth mentioning that this flow can't be used by an actual end user (already mentioned in [https://developer.okta.com/docs/concepts/oauth-openid/#client-credentials-flow](https://developer.okta.com/docs/concepts/oauth-openid/#client-credentials-flow))

For Resource Owner Password Flow, could we make the wording similar to other flows (you pass it to your [authorization server's](https://developer.okta.com/docs/concepts/auth-servers/) /token endpoint. If you are using the default Okta authorization server, then your request would look something like this:)?

susanharper-okta:
@codepadma - I have updated the PR with your comments. Can you please take a look? Thank you.

packages/@okta/vuepress-site/docs/guides/implement-password/use-flow/index.md
```diff
@@ -2,9 +2,7 @@
 title: Use the Resource Owner Password Flow
 ---
 
-Before you can begin this flow, you will have to collect the user's password in a manner of your choosing.
-
-Once you have collected the credentials, all that is required is a single API call to the `/token` endpoint:
+Before you can begin this flow, collect the user's password in a manner of your choosing. Once you have collected the credentials, these are then passed through Basic Auth in the request to your [Custom Authorization Server's](/docs/concepts/auth-servers/#custom-authorization-server) `/token` endpoint:
```
codepadma, line 5: Before you can begin this flow, you will have to collect the user's password in a manner of your choosing. Once you have collected the credentials, all that is required is a single API call to the [authorization server's](https://developer.okta.com/docs/concepts/auth-servers/) `/token` endpoint. If you are using the default Okta authorization server, then your request would look something like this:

codepadma (changes requested):

packages/@okta/vuepress-site/docs/guides/implement-password/use-flow/index.md
susanharper-okta: I have updated - I also have changed "default Okta authorization server" to default Custom Authorization Server" as there is ongoing confusion with that - so typically clarify "default Custom Authorization Server".

PR #1683, opened 2020-12-10
Fix Incorrect Link Specify User Type of a New User

colinwang-okta:
## Description:
- **What's changed?** Fix link on "full replace of an existing user" under the "Specify the User Type of a new user" section
- **Is this PR related to a Monolith release?** no

### DIFF
--- a/packages/@okta/vuepress-site/docs/reference/api/user-types/index.md
+++ b/packages/@okta/vuepress-site/docs/reference/api/user-types/index.md
@@ -308,7 +308,7 @@ HTTP/1.1 204 No Content
 
 ## Specify the User Type of a new user
 
-The [Create User](/docs/reference/api/users/#create-user-with-non-default-user-type) operation accepts a type specification as part of the request body. The specification is a map, but currently the only key permitted is `id`. The type specification is also added to the [User object](/docs/reference/api/users/#user-object), but after user creation the type may only be updated by an administrator on a [full replace of an existing user](/docs/reference/api/user-types/#update-user-type), not a partial update.
+The [Create User](/docs/reference/api/users/#create-user-with-non-default-user-type) operation accepts a type specification as part of the request body. The specification is a map, but currently the only key permitted is `id`. The type specification is also added to the [User object](/docs/reference/api/users/#user-object), but after user creation the type may only be updated by an administrator on a [full replace of an existing user](/docs/reference/api/users/#update-user), not a partial update.
 
 ##### Example
 
--- a/packages/@okta/vuepress-site/docs/reference/api/users/index.md
+++ b/packages/@okta/vuepress-site/docs/reference/api/users/index.md
@@ -1624,7 +1624,7 @@ in the request is deleted.
 
 `profile` and `credentials` can be updated independently or together with a single request.
 
->**Note:** Currently, the User Type of a user can only be changed via a full replacement [PUT operation](/docs/reference/api/user-types/#update-user-type). If the Request Parameters of a partial update include the `type` element from the [User object](#user-object), the value must match the existing type of the user. Only administrators are permitted to change the user type of a user; end users are not allowed to change their own user type.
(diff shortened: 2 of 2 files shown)


chenningzhang-okta (approved):
LGTM

packages/@okta/vuepress-site/docs/reference/api/user-types/index.md
```diff
@@ -308,7 +308,7 @@ HTTP/1.1 204 No Content
 
 ## Specify the User Type of a new user
 
-The [Create User](/docs/reference/api/users/#create-user-with-non-default-user-type) operation accepts a type specification as part of the request body. The specification is a map, but currently the only key permitted is `id`. The type specification is also added to the [User object](/docs/reference/api/users/#user-object), but after user creation the type may only be updated by an administrator on a [full replace of an existing user](/docs/reference/api/user-types/#update-user-type), not a partial update.
+The [Create User](/docs/reference/api/users/#create-user-with-non-default-user-type) operation accepts a type specification as part of the request body. The specification is a map, but currently the only key permitted is `id`. The type specification is also added to the [User object](/docs/reference/api/users/#user-object), but after user creation the type may only be updated by an administrator on a [full replace of an existing user](/docs/reference/api/users/#update-user), not a partial update.
```
IanHakes-okta, line 311: The link works, but I'm not clear on **why** you're linking away from the Update User Type description. The previous destination link (Update User Type) explains that that a PUT does a full replace operation on the user type; the new destination link (Update User) isn't clear up front about how to do that full replace operation to make that change - in fact, the note at the bottom sends them right back to the "full replacement PUT operation" on the Update User Type. I think you need to add some text to explain exactly what the dev needs to do.

colinwang-okta: The existing link to the Update User Type description is not correct. That's where you would change things about the User Type itself not the User Type for a specific user. You are correct that the link in the bottom of the user update section is also incorrect, so thank you for pointing that out, and I've removed the link there. I'm not actually part of the UD team. I just want to fix the mistake in the docs quickly. Could we at least fix the broken links here with my PR, and file a ticket with the ud team, if you feel like there is additional work to be done on the description?

IanHakes-okta: OK - can you open the JIRA against the UD team to fix please? I will approve these fixes.

IanHakes-okta (approved):

packages/@okta/vuepress-site/docs/reference/api/user-types/index.md
colinwang-okta: I'm actually not sure what I would change about the existing description, so I don't actually have anything to file against the UD team. If you have any specific suggestions though you can follow up with them.

packages/@okta/vuepress-site/docs/reference/api/users/index.md
```diff
@@ -1624,7 +1624,7 @@ in the request is deleted.
 
 `profile` and `credentials` can be updated independently or together with a single request.
 
->**Note:** Currently, the User Type of a user can only be changed via a full replacement [PUT operation](/docs/reference/api/user-types/#update-user-type). If the Request Parameters of a partial update include the `type` element from the [User object](#user-object), the value must match the existing type of the user. Only administrators are permitted to change the user type of a user; end users are not allowed to change their own user type.
+>**Note:** Currently, the User Type of a user can only be changed via a full replacement PUT operation. If the Request Parameters of a partial update include the `type` element from the [User object](#user-object), the value must match the existing type of the user. Only administrators are permitted to change the user type of a user; end users are not allowed to change their own user type.
```
chenningzhang-okta, line 1627: I think the link was removed b/c this note is part of the user update section already. So no need to add a link to the section itself. This change looks good to me as well.

chenningzhang-okta (approved):

PR #1699, opened 2020-12-14
Add redirection for deprecated links

haishengwu-okta:
## Description:
- **What's changed?** Update redirections what would be used after migration 

### Resolves:

* [OKTA-346418](https://oktainc.atlassian.net/browse/OKTA-346418)

### DIFF
--- a/packages/@okta/vuepress-site/conductor.yml
+++ b/packages/@okta/vuepress-site/conductor.yml
@@ -1991,3 +1991,11 @@ redirects:
     to: /docs/guides/style-the-widget/customization-examples/
   - from: /docs/guides/custom-hosted-signin/customization-examples/index.html
     to: /docs/guides/style-the-widget/customization-examples/
+  - from: /contact-sales-enterprise
+    to: https://www.okta.com/contact-sales/
+  - from: /contact-us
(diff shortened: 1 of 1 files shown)


albertchen-okta:
@hamzaessaoui-okta fyi - if we want to add any UTM values to these redirects, this is the PR to do it in

francoislehoux-okta (approved):

shawyu-okta (approved):

hamzaessaoui-okta:
@francoislehoux-okta  @shawyu-okta @albertchen-okta will the redirect preserve query parameters:
ex: https://developer.okta.com/contact-sales-enterprise?utm_source=google&utm_term=emea should redirect to https://www.okta.com/contact-sales/?utm_source=google&utm_term=emea

UTMs are not unique, the values will be different for most requests

shawyu-okta:
@hamzaessaoui-okta @albertchen-okta -- I believe in Slack we discussed that the redirect will NOT preserve query parameters (conductor limitation) but we will ask that the origin of the URLs (Sales) update their links. The redirects are just a convenience to avoid broken links but all the links should be updated to avoid needing a redirect in the first place.

PR #1731, opened 2020-12-22
Remove X-Device-Fingerprint from docs

Tennyx:
This header was deprecated on 12/7/2020 I believe, should probably remove it from the docs.

## Description:
- **What's changed?** 

### Resolves:


jakubvul-okta:
Thanks for catching this @Tennyx ! It's not gone, but deprecated, so we want to keep this in our documentation for our older customers. I've created a PR that adds the "Deprecated" tag here: https://github.com/okta/okta-developer-docs/pull/1821

PR #1735, opened 2020-12-23
Adds new navbar structure

volodymyrmakarchuk-okta:
## Description:
- **What's changed?** 
  * Hides breadcrumbs in redesign
  * Breadcrumbs still shown on mobile viewports in redesign
  * Navigation contains all top level items, not just docs items
  * "Docs" breadcrumb is not a link anymore in redesign

### Resolves:

* [OKTA-351798](https://oktainc.atlassian.net/browse/OKTA-351798)


packages/@okta/vuepress-theme-prose/components/SidebarItem.vue
```diff
@@ -19,6 +19,7 @@
             </div>
         </div>
 
+
```
francoislehoux-okta, line 22: Remove `newline`

packages/@okta/vuepress-theme-prose/index.js
```diff
@@ -16,7 +16,7 @@ module.exports = (options, ctx) => {
       // Set to true to enable the redesign in development
       const redesign = process.env.REDESIGN || false;
       // Respect the redesign flag unless this is a production build
-      $page.redesign = !ctx.isProd && redesign;
+      $page.redesign = true //!ctx.isProd && redesign;
```
francoislehoux-okta, line 19: Note for removal

packages/@okta/vuepress-site/.vuepress/config.js
```diff
@@ -2,6 +2,7 @@ const guidesInfo = require('./scripts/build-guides-info');
 const findLatestWidgetVersion = require('./scripts/findLatestWidgetVersion');
 const convertReplacementStrings = require('./scripts/convert-replacement-strings');
 const signInWidgetMajorVersion = 5;
+const redesign = true; //!!process.env.REDESIGN;
```
francoislehoux-okta, line 5: Note for removal

packages/@okta/vuepress-site/.vuepress/config.js
```diff
@@ -132,10 +133,10 @@ module.exports = {
     ],
 
     sidebars: {
-      codePages: require('./nav/codePages'),
-      reference: require('./nav/reference'),
-      guides: require('./nav/guides'),
-      concepts: require('./nav/concepts')
+      codePages: redesign ? require('./nav/newNav/languagesSdk') : require('./nav/codePages'),
```
francoislehoux-okta, line 136: Can we keep it to `redesign` here instead of `newNav`? Will be much easier to find and replace in a couple of weeks.

packages/@okta/vuepress-site/.vuepress/nav/newNav/guides.js
```diff
@@ -0,0 +1,232 @@
...
+  {
+    title: "Guides",
+    subLinks: [
+      { title: 'Guides overview', path: '/docs/guides/'},
+      {
+        title: "Basics",
+        subLinks: [
+          {
+            title: "Quickstart: Signing in your first user",
+            guideName: "quickstart"
```
francoislehoux-okta, line 13: Not familiar with this so asking, any reason this file uses guideNames but all 3 other uses paths?

volodymyrmakarchuk-okta: yes it's something that was done previously by other developers, guides include some additional logic so I didn't want to touch it

packages/@okta/vuepress-site/.vuepress/config.js
shawyu-okta, line 5: Is there any way to pull this from the existing FF location? If not, we should refactor both into a single place for simplicity -- best not to have this appearing in multiple locations.

packages/@okta/vuepress-theme-prose/components/Sidebar.vue
```diff
@@ -1,43 +1,54 @@
...
+      navBarItems: []
     };
   },
   mounted() {
-    if(!this.appContext.isInMobileViewport) {
+    if (!this.appContext.isInMobileViewport) {
       this.handleScroll();
       window.addEventListener("scroll", this.handleScroll);
     }
+    this.navBarItems = this.navigation;
```
shawyu-okta, line 47: Is this necessary, or can we we directly access the computed `navigation`?

packages/@okta/vuepress-site/.vuepress/config.js
volodymyrmakarchuk-okta: yep, because it is config.js file `$page` and etc seems not available here

shawyu-okta: I'm wondering if we can store the sidebar items as two separate config items and then switch on it in the component so we can use the `$page.redesign` FF there to differentiate, rather than do it here.

packages/@okta/vuepress-theme-prose/components/Sidebar.vue
volodymyrmakarchuk-okta: if we will directly access navigation it will be triggering each time we click on navbar item

packages/@okta/vuepress-site/.vuepress/config.js
volodymyrmakarchuk-okta: and from where config.js can be load programmatically?

shawyu-okta: I think the answer is that we should treat config as static -- so put them both here and pick which one you want in the component instead.

packages/@okta/vuepress-theme-prose/components/Sidebar.vue
shawyu-okta: Ah I see. Then we can directly inline the computation method in here rather than do a reassignment, e.g.
```js
this.navBarItems = (this.getNavigation() || []).map(/* the rest of the code */;
```

packages/@okta/vuepress-theme-prose/components/Sidebar.vue
```diff
@@ -1,43 +1,54 @@
 <template>
   <aside class="landing-navigation">
     <ul class="landing">
-      <SidebarItem v-for="link in navigation" :key="link.title" :link="link" />
+      <ul class="sections">
+        <SidebarItem
+          v-for="link in navBarItems"
```
oleksandrpravosudko-okta, line 6: this might be breaking active sidebar item highlight on URL change with stack selector (e.g. https://developer.okta.com/docs/guides/sign-into-spa/react/handle-callback/)

oleksandrpravosudko-okta: it looks like this is happening b/c `navBarItems` is not getting updated on Sidebar update

packages/@okta/vuepress-site/.vuepress/config.js
volodymyrmakarchuk-okta: There was a suggestion to move this feature flag to config.js thus it will be available in vue components via `his.$site.themeConfig.redesign` and also it will be possible to use it in config.js as well

volodymyrmakarchuk-okta: but it will require some refactoring after

packages/@okta/vuepress-theme-prose/components/Sidebar.vue
volodymyrmakarchuk-okta: Seems @ pradeepdewda-okta comment makes sense. Such changes will break some logic related to switching between vue / react / angular documentation by using [select element] (https://developer.okta.com/docs/guides/sign-into-spa/angular/add-signin-button /). So here when we switching between frameworks rout path is changing and navbar should be rerendered not lose the selected item. So probably a current solution with permanent rendering because of directly using `navigation` is not perfect but it will not break anything. And after that, we can create a task for using a more elegant solution in case of switching between different frameworks in documentation pages

shawyu-okta: Ok, we can leave this as-is if it's needed for functionality

francoislehoux-okta:
Do you have a working `netlify` build you could share?

volodymyrmakarchuk-okta:
> Do you have a working `netlify` build you could share?

https://5fe4c066d3fff9af92fec56b--reverent-murdock-829d24.netlify.app/docs/guides/

packages/@okta/vuepress-theme-prose/components/Sidebar.vue
```diff
@@ -1,43 +1,65 @@
...
   </aside>
 </template>
 
 <script>
 import _ from "lodash";
-import { getGuidesInfo, guideFromPath } from "../util/guides";  
+import { getGuidesInfo, guideFromPath } from "../util/guides";
+import {
+  concepts as conceptsRedesign,
+  guides as gidesRedesign,
```
shawyu-okta, line 20:
```suggestion
  guides as guidesRedesign,
```

packages/@okta/vuepress-theme-prose/components/Sidebar.vue
```diff
@@ -94,27 +126,32 @@ export default {
           .value();
       }
     },
-    getGuides() {
+    getGuides(guidesRedesign) {
```
shawyu-okta, line 129:
```suggestion
    getGuides(overrides) {
```

shawyu-okta (approved):

francoislehoux-okta (approved):
LGTM! :shipit:

PR #1756, opened 2021-01-02
Consent section & minor singn up enhancement

volodymyrmakarchuk-okta:
## Description:
- **What's changed?** 

### Resolves:

* [OKTA-355579](https://oktainc.atlassian.net/browse/OKTA-355579)
* [OKTA-355582](https://oktainc.atlassian.net/browse/OKTA-355582)

### DIFF
--- a/packages/@okta/vuepress-theme-prose/global-components/SignUp.vue
+++ b/packages/@okta/vuepress-theme-prose/global-components/SignUp.vue
@@ -1,7 +1,11 @@
 <template>
   <div class="signup">
     <div class="signup--form">
-      <form @submit="submitForm" method="POST" action="https://developer.okta.com/developer/signup/">
+      <form
+        @submit="submitForm"
+        method="POST"
+        action="https://developer.okta.com/developer/signup/"
+      >
         <div class="row">
           <label class="field-wrapper" for="email">
             Email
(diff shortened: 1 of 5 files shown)


packages/@okta/vuepress-theme-prose/global-components/SignUp.vue
```diff
@@ -255,9 +299,27 @@ export default {
...
 
       if (this.validationService.isValidForm()) {
-       // make api call
+        // make api call
+      }
+    },
+
+    showConsentSection(country) {
+      // Show consent section while selecting country.
+      this.displayConsent = true;
```
shawyu-okta, line 311: What is the purpose of this variable? It seems like it only gets set to `true` but never gets toggled back anywhere else?

packages/@okta/vuepress-theme-prose/global-components/SignUp.vue
```diff
@@ -255,9 +299,27 @@ export default {
...
+      // Hide consent opt checkbox.
+      this.displayAgree = false;
+      this.form.consentAgree.hidden = true;
+      this.validationService.resetFormField("consentAgree", {
+        reset: true,
+        value: false
+      });
+
+      if (GDPR_COUNTRIES.indexOf(country) !== -1) {
```
shawyu-okta, line 320: Can we use `.contains`? It's more appropriate.

packages/@okta/vuepress-theme-prose/global-components/SignUp.vue
```diff
@@ -26,52 +30,52 @@
...
+                placeholder="First Name"
+                v-model="form.firstName.value"
+                :class="{ error: !form.firstName.isValid }"
+                @blur="validationService.checkFormInput('firstName')"
+              />
+              <ul class="error-color" v-if="form.firstName.errorList.length">
+                <li
+                  class="error-color"
+                  v-for="error in form.firstName.errorList"
+                  v-bind:key="error.length * Math.random()"
```
shawyu-okta, line 49: I know this is just an indent change but I noticed this-- `v-for` also exposes `index` as a second loop parameter, so we may want to use that:
```js
v-for="(error, index) in form.firstName.errorList"
```

shawyu-okta: If you make that change, also do the same below for the same scenario.

volodymyrmakarchuk-okta: and where we want to use index after?

packages/@okta/vuepress-theme-prose/global-components/SignUp.vue
volodymyrmakarchuk-okta: as soon as user will select any country consent block becomes visible and it will be visible for all countries. At the same time, opt checkbox will be visible only for gdpr countries, you can check the current implementation of this part on prod version

packages/@okta/vuepress-theme-prose/global-components/SignUp.vue
shawyu-okta: Use the index instead of `error.length * Math.random()` in the `key` field

packages/@okta/vuepress-theme-prose/global-components/SignUp.vue
shawyu-okta: Ah, okay, thank you for clarifying.

packages/@okta/vuepress-theme-prose/global-components/SignUp.vue
shawyu-okta: Never mind, it looks like vuepress version we use only has `core-js` 2.x which at the time only had `.includes` as a Stage 4 proposal so it's probably not on by default. We'll just keep `indexOf`.

shawyu-okta (approved):

PR #1773, opened 2021-01-08
Fix mistake on user lifecycle update doc.

chenningzhang-okta:
## Description:
- /api/v1/users/{id}/lifecycle/activate can be performed on users with "DEPROVISIONED" status.

@nghadge-okta @okta/ud-core

### DIFF
--- a/packages/@okta/vuepress-site/docs/reference/api/users/index.md
+++ b/packages/@okta/vuepress-site/docs/reference/api/users/index.md
@@ -2141,7 +2141,7 @@ For example, you can't unlock a user that is `ACTIVE`.
 
 Activates a user
 
-This operation can only be performed on users with a `STAGED` status.  Activation of a user is an asynchronous operation.
+This operation can only be performed on users with a `STAGED` or `DEPROVISIONED` status.  Activation of a user is an asynchronous operation.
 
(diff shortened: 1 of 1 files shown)


arvindkrishnakumar-okta (approved):

packages/@okta/vuepress-site/docs/reference/api/users/index.md
```diff
@@ -2141,7 +2141,7 @@ For example, you can't unlock a user that is `ACTIVE`.
 
 Activates a user
 
-This operation can only be performed on users with a `STAGED` status.  Activation of a user is an asynchronous operation.
+This operation can only be performed on users with a `STAGED` or `DEPROVISIONED` status.  Activation of a user is an asynchronous operation.
```
nghadge-okta, line 2144: "Activation of a user is an asynchronous operation" is this true? because I thought from API its is async only when they pass a header otherwise its still sync process.

donwoods-okta: Nope.  The `activate` and `reactivate` endpoints do not accept the special header for async.  I chased through the flow and found that they set the status to `ACTIVATING`, which does sound like the rest of the work is unconditionally asynchronous.

nghadge-okta (approved):

PR #1789, opened 2021-01-14
Adds URI configurations for login and signup

shawyu-okta:
## Description:
- **What's changed?** 
  * Adds URI configurations for login and signup targets
  * Currently pointed at trex URIs. Will be updated to production in go-live PR
  * Removes CAPTCHA enforcement in non-production scenarios

### Resolves:

* [OKTA-360692](https://oktainc.atlassian.net/browse/OKTA-360692)


packages/@okta/vuepress-theme-prose/index.js
```diff
@@ -14,9 +14,9 @@ module.exports = (options, ctx) => {
       $page.fullHeaders = resolveHeaders($page);
 
       // Set to true to enable the redesign in development
-      const redesign = process.env.REDESIGN || false;
+      //const redesign = process.env.REDESIGN || false;
       // Respect the redesign flag unless this is a production build
-      $page.redesign = !ctx.isProd && redesign;
+      $page.redesign = true;
```
francoislehoux-okta, line 19: Note

kamaldeepcheema-okta: just for my understanding, does this mean we're ignoring the redesign flag here from now on?

shawyu-okta: Ah no, sorry this is just turned on in the PR for the Netlify deploy. Will be turning it off before merging @kamaldeepcheema-okta

kamaldeepcheema-okta: ah ok, thanks for the clarification!

packages/@okta/vuepress-theme-prose/global-components/SignUp.vue
```diff
@@ -300,9 +307,25 @@ export default {
...
+          userProfile: {
+            email: this.form.email.value,
+            firstName: this.form.firstName.value,
+            lastName: this.form.lastName.value,
+            country: this.form.country.value,
+            state: this.form.state.value,
+            emailOptInC: this.form.consentAgree.value,
+            captchaResponse: this.form.captcha.value,
+          },
+        };
```
volodymyrmakarchuk-okta, line 326: just a minor note, in the next commits, probably will nice to wrap this part of code in some specific function and not to store it here

volodymyrmakarchuk-okta (approved):

packages/@okta/vuepress-theme-prose/global-components/Login.vue
```diff
@@ -52,5 +45,17 @@
...
+export default {
+  computed: {
+    uris() {
+      const { uris } = this.$site.themeConfig;
+
+      return {
+        email: "https://login.okta.com/",
+        github: `${uris.baseUri}/sso/idps/${uris.idps.github}`,
+        google: `${uris.baseUri}/sso/idps/${uris.idps.google}`,
+      };
```
francoislehoux-okta, line 57: Can we make a helper that builds these? Seems like they're the same in both login and signup

shawyu-okta: I had put it off since it's only used two places but I think it's worthwhile to avoid making a mistake changing one URI but not the other. I've added a commit to address this

francoislehoux-okta (approved):
LGTM

PR #1799, opened 2021-01-15
Correctly renew Coveo's access token

francoislehoux-okta:
## Description:
- **What's changed?**
- Update coveo to a newer version which has a working renewal callback
- Locally invalidate the token to prevent UI from flickering

### Resolves:

* [OKTA-361551](https://oktainc.atlassian.net/browse/OKTA-361551)

### DIFF
--- a/packages/@okta/vuepress-theme-prose/util/coveo/endpoint.js
+++ b/packages/@okta/vuepress-theme-prose/util/coveo/endpoint.js
@@ -1,27 +1,43 @@
 import axios from "axios";
+import jwt_decode from "jwt-decode";
+import moment from "moment";
 
 const TOKEN_ENDPOINT =
   "https://wt1ugse0be.execute-api.us-west-2.amazonaws.com/prod/token?site=developer";
 const COVEO_PIPELINE = "oktaproduction9ounvcxa";
 const COVEO_ENDPOINT = "https://platform.cloud.coveo.com/rest/search";
 
-const _getToken = async () => {
-  const {
-    data: { token }
-  } = await axios.get(TOKEN_ENDPOINT);
+const _getToken = () => {
+  const token = localStorage.getItem("coveo_token");
+  if (!token) return;
 
-  localStorage.setItem("coveo_token", token);
-  return token;
+  // Invalidate token if about to expire
+  const decoded = jwt_decode(token);
+  const now = moment();
+  const expWithBuffer = moment.unix(decoded.exp).subtract(5, "minutes");
+  if (now.isAfter(moment(expWithBuffer).subtract(5, "minutes"))) {
+    localStorage.removeItem("coveo_token");
+    return;
+  } else {
+    return token;
+  }
+};
+
+const _renewToken = () => {
+  return axios.get(TOKEN_ENDPOINT).then(({ data: { token } }) => {
+    localStorage.setItem("coveo_token", token);
+    return token;
+  });
 };
 
 const configureCoveoEndpoint = async () => {
-  const token = localStorage.getItem("coveo_token") || (await _getToken());
+  const token = _getToken() || (await _renewToken());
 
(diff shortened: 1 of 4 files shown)


packages/@okta/vuepress-theme-prose/util/coveo/endpoint.js
```diff
@@ -1,27 +1,43 @@
...
-  } = await axios.get(TOKEN_ENDPOINT);
+const _getToken = () => {
+  const token = localStorage.getItem("coveo_token");
+  if (!token) return;
 
-  localStorage.setItem("coveo_token", token);
-  return token;
+  // Invalidate token if expired
+  const decoded = jwt_decode(token);
+  const now = moment(Date.now() / 1000);
```
shawyu-okta, line 16: Is this the same as `moment()`?

packages/@okta/vuepress-theme-prose/util/coveo/endpoint.js
```diff
@@ -1,27 +1,43 @@
...
+  const exp = moment(decoded.exp);
+  if (exp.isSameOrBefore(now)) {
```
shawyu-okta, line 18: Can this be written as `if ( now isafter exp )`? Think it would be easier to parse

packages/@okta/vuepress-theme-prose/index.js
```diff
@@ -14,9 +14,10 @@ module.exports = (options, ctx) => {
       $page.fullHeaders = resolveHeaders($page);
 
       // Set to true to enable the redesign in development
-      const redesign = process.env.REDESIGN || false;
+      // const redesign = process.env.REDESIGN || false;
       // Respect the redesign flag unless this is a production build
-      $page.redesign = !ctx.isProd && redesign;
+      // $page.redesign = !ctx.isProd && redesign;
+      $page.redesign = true;
```
francoislehoux-okta, line 20: Cleanup before merge

packages/@okta/vuepress-theme-prose/util/coveo/endpoint.js
shawyu-okta: Also, should we get a new token anyways if we are close (say, 5 minutes) from expiration?

packages/@okta/vuepress-theme-prose/index.js
shawyu-okta, line 20: reminder to revert

shawyu-okta (approved):
Couple of suggestions/clarifications but overall logic seems sound, won't block it

packages/@okta/vuepress-theme-prose/util/coveo/endpoint.js
francoislehoux-okta: Was converting this one to match `exp`'s format but now doing it the other way around with `moment.unix(exp)` which does `* 1000` internally

packages/@okta/vuepress-theme-prose/util/coveo/endpoint.js
```diff
@@ -1,27 +1,43 @@
...
+  const token = localStorage.getItem("coveo_token");
+  if (!token) return;
 
-  localStorage.setItem("coveo_token", token);
-  return token;
+  // Invalidate token if about to expire
+  const decoded = jwt_decode(token);
+  const now = moment();
+  const expWithBuffer = moment.unix(decoded.exp).subtract(5, "minutes");
+  if (now.isAfter(moment(expWithBuffer).subtract(5, "minutes"))) {
```
shawyu-okta, line 18: Looks like we are actually doing a buffer of 10 minutes, not sure if it's what you intended but doesn't hurt much either.

francoislehoux-okta: Whoops, good catch, doesn't hurt but does not look good ^^

PR #1821, opened 2021-01-20
Adding deprecated note for device fingerprint header

jakubvul-okta:
## Description:
- **What's changed?** Adding note that this header is deprecated

**Screenshot:**
Screen Shot 2021-01-20 at 1 13 55 PM

### Resolves:

#1731

### DIFF
--- a/packages/@okta/vuepress-site/docs/reference/api-overview/index.md
+++ b/packages/@okta/vuepress-site/docs/reference/api-overview/index.md
@@ -119,6 +119,8 @@ The `Accept-Language` HTTP header advertises which languages the client is able 
 
 ## Device Fingerprint
 
+<ApiLifecycle access="deprecated" />
+
 The `X-Device-Fingerprint` HTTP header supplies the device fingerprint used in an authentication request.
 
 ## Errors


packages/@okta/vuepress-site/docs/reference/api-overview/index.md
```diff
@@ -119,6 +119,8 @@ The `Accept-Language` HTTP header advertises which languages the client is able
 
 ## Device Fingerprint
 
+<ApiLifecycle access="deprecated" />
```
srinivasanagandla-okta, line 122: Instead of marking this as deprecated, if device fingerprint is not used in sign-on policy evaluation we should delete the highlighted sentence in the following section.

Screen Shot 2021-01-20 at 11 19 23 AM

My understanding is that we still use X-Device-Fingerprint for trusted apps but the usage by Okta is narrowed down to just sending new device email notifications.

jakubvul-okta:
@Tennyx because this is only deprecated for some orgs, we're actually going to leave the documentation as is for now while we sort out deployment. Sorry about the confusion!

PR #1855, opened 2021-01-27
Update index.md

aidankinzett:
After going through the verification process with ZeroSSL it didn't work because Okta wont accept that format of certificate. Don't tell people to go and use a website that wont work with your product.

## Description:
- Added the note from https://support.okta.com/help/s/article/Custom-URL-certificate-upload-is-failing-when-generated-by-ZeroSSL?language=en_US so that other developers dont waste time with ZeroSSL


albertchen-okta:
@aidankinzett Sorry about that confusion, that is our fault. I have removed references to ZeroSSL in the dev doc guide. Looks like that ACME option could work, but probably causes more unnecessary confusion to hunt down those details. 

Also just an fyi, we are working on a option to allow ACME through Let's Encrypt directly from our Admin Console UI, so stay tuned for a simpler world.

PR #1865, opened 2021-01-28
Uses CDN to load OktaSignIn

shawyu-okta:
## Description:
- **What's changed?** 
  * Loads SIW from Okta CDN instead of from installed NPM module

### Resolves:

* [OKTA-365567](https://oktainc.atlassian.net/browse/OKTA-365567)

### DIFF
--- a/packages/@okta/vuepress-site/.vuepress/config.js
+++ b/packages/@okta/vuepress-site/.vuepress/config.js
@@ -23,6 +23,12 @@ module.exports = {
     ['meta', { 'http-equiv': 'XA-UA-Compatible', content: 'IE=edge'}],
     ['meta', { name: 'viewport', content: 'width=device-width, initial-scale=1.0'}],
 
+    /*
+     * Okta sign-in widget
+     */
+    ['link', { href: 'https://global.oktacdn.com/okta-signin-widget/5.2.0/css/okta-sign-in.min.css', type: 'text/css', rel: 'stylesheet'}],
+    ['script', { src: 'https://global.oktacdn.com/okta-signin-widget/5.2.0/js/okta-sign-in.min.js', type: 'text/javascript'}],
+
     /**
      * Header scripts for typekit, GA, GTM (WIP)
      */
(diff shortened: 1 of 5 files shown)


packages/@okta/vuepress-site/.vuepress/config.js
```diff
@@ -23,6 +23,12 @@ module.exports = {
     ['meta', { 'http-equiv': 'XA-UA-Compatible', content: 'IE=edge'}],
     ['meta', { name: 'viewport', content: 'width=device-width, initial-scale=1.0'}],
 
+    /*
+     * Okta sign-in widget
+     */
+    ['link', { href: 'https://global.oktacdn.com/okta-signin-widget/5.2.0/css/okta-sign-in.min.css', type: 'text/css', rel: 'stylesheet'}],
```
haishengwu-okta, line 29: - Create a variable for the version number to allow easy upgrade?
- Is the comment in the editor auto generated?. Are we able to update that as well?

`// Uses okta-signin-widget version 5.2.0`

francoislehoux-okta:
I noticed the CodeMirror stuff also has trouble loading. Any way we could use a CDN as well ?

shawyu-okta:
I'm not sure if it's kosher to use code hosted by a 3rd party mirror for CodeMirror but if we somehow hosted it on our own CDN we could.

packages/@okta/vuepress-site/.vuepress/config.js
shawyu-okta: I just found this widget version script that exists for updating the documentation automatically so I'm going to try to leverage it here

shawyu-okta: Updated it -- this pulls from the same version that is shown on our docs pages now and it works really well

francoislehoux-okta:
I believe we require Coveo from a CDN with SRI enabled and the integrity hash. Not sure we can find a version of code mirror with it.

I have a feeling this will get the SIW to render, but the code mirror part will still fail from time to time.

francoislehoux-okta (approved):
LGTM

PR #1873, opened 2021-01-31
Fixing sitemap link 

kamaldeepcheema-okta:
## Description:
- **What's changed?** 
> Fixing the sitemap link. Previously we were displaying the sitemap under the 'docs-sitemap' alias.

> No

### Resolves:

* [OKTA-362166](https://oktainc.atlassian.net/browse/OKTA-362166)

**Netlify Build

https://6016f373b01495f73debdbe0--reverent-murdock-829d24.netlify.app/sitemap.xml

### DIFF
--- a/packages/@okta/vuepress-site/.vuepress/config.js
+++ b/packages/@okta/vuepress-site/.vuepress/config.js
@@ -326,14 +326,14 @@ module.exports = {
   plugins: {
     'sitemap': {
       hostname: 'https://developer.okta.com',
-      outFile: 'sitemap.xml',
+      outFile: 'docs-sitemap.xml',
       exclude: [
         '/test_page/'
       ]
-    },
+    }, 
     'robots': {
       host: 'https://developer.okta.com',
-      sitemap: '/sitemap.xml',
+      sitemap: '/docs-sitemap.xml',
       policies: [
         {
           userAgent: '*',
--- a/packages/@okta/vuepress-site/.vuepress/config.js
+++ b/packages/@okta/vuepress-site/.vuepress/config.js
@@ -333,7 +333,7 @@ module.exports = {
     },
     'robots': {
       host: 'https://developer.okta.com',
-      sitemap: '/docs-sitemap.xml',
+      sitemap: '/sitemap_index.xml',
       policies: [
         {
           userAgent: '*',
--- a/packages/@okta/vuepress-site/.vuepress/config.js
+++ b/packages/@okta/vuepress-site/.vuepress/config.js
@@ -330,7 +330,7 @@ module.exports = {
       exclude: [
         '/test_page/'
       ]
-    }, 
+    },
     'robots': {
       host: 'https://developer.okta.com',
       sitemap: '/docs-sitemap.xml',
--- a/packages/@okta/vuepress-site/.vuepress/config.js
+++ b/packages/@okta/vuepress-site/.vuepress/config.js
@@ -326,14 +326,14 @@ module.exports = {
   plugins: {
     'sitemap': {
       hostname: 'https://developer.okta.com',
-      outFile: 'docs-sitemap.xml',
+      outFile: 'sitemap.xml',
       exclude: [
         '/test_page/'
       ]
     },
     'robots': {
       host: 'https://developer.okta.com',
-      sitemap: '/docs-sitemap.xml',
+      sitemap: '/sitemap.xml',
       policies: [
         {
           userAgent: '*',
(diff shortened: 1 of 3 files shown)


francoislehoux-okta (commented):
Please verify with Hamza, this will most likely break `Coveo`. It expects to have a a `docs-sitemap` file for docs related pages, and other for the other categories.

I think you need to keep that one sitemap as is and generate more for non doc related pages.

hamzaessaoui-okta:
@kamaldeepcheema-okta @francoislehoux-okta docs can't use /sitemap.xml since it's being used by blog.
Is there a reason we want to rename it?

In all cases, this wouldn't break Coveo since it uses https://developer.okta.com/sitemap_index.xml. If we do rename the docs sitemap, we need to update the sitemap_index file.

francoislehoux-okta:
> @kamaldeepcheema-okta @francoislehoux-okta docs can't use /sitemap.xml since it's being used by blog.
> Is there a reason we want to rename it?
> 
> In all cases, this wouldn't break Coveo since it uses https://developer.okta.com/sitemap_index.xml. If we do rename the docs sitemap, we need to update the sitemap_index file.

This is what I was looking for. To sum it up:
- `sitemap.xml` is the blog's sitemap
- `pages-sitemap.xml` is the current drupal's sitemap (developer pages, non-docs)
- `docs-sitemap.xml` is the current vuepress' sitemap (docs pages)

Short term, drupal will be going out and most of these pages have been migrated over to vuepress. If I understand correctly, we need to have a `pages-sitemap.xml` generated on the vuepress site for any page that is not documentation, so pricing, oin, etc? @hamzaessaoui-okta

kamaldeepcheema-okta:
> > @kamaldeepcheema-okta @francoislehoux-okta docs can't use /sitemap.xml since it's being used by blog.
> > Is there a reason we want to rename it?
> > In all cases, this wouldn't break Coveo since it uses https://developer.okta.com/sitemap_index.xml. If we do rename the docs sitemap, we need to update the sitemap_index file.
> 
> This is what I was looking for. To sum it up:
> 
> * `sitemap.xml` is the blog's sitemap
> * `pages-sitemap.xml` is the current drupal's sitemap (developer pages, non-docs)
> * `docs-sitemap.xml` is the current vuepress' sitemap (docs pages)
> 
> Short term, drupal will be going out and most of these pages have been migrated over to vuepress. If I understand correctly, we need to have a `pages-sitemap.xml` generated on the vuepress site for any page that is not documentation, so pricing, oin, etc? @hamzaessaoui-okta
> 
> Where is sitemap_index.xml generated/hosted from? @hamzaessaoui-okta

I've compared the current https://developer.okta.com/docs-sitemap.xml VS https://developer.okta.com/pages-sitemap.xml and I can see that our docs-sitemap.xml already has a lot of the files listed in pages-sitemap.xml. The ones that it does not are paths that don't exist anymore. So in fact it looks like the pages-sitemap.xml file is outdated and our docs-sitemap.xml is the more current version.

In regards to the contents of the blog's sitemap.xml - that isn't included at all in our docs-sitemap.xml.

hamzaessaoui-okta:
@francoislehoux-okta @kamaldeepcheema-okta if the goal is to take over the sitemap, I would recommend:
- creating a static file sitemap_index.xml with two links (docs sitemap - the name doesn't matter) + blog sitemap - sitemap.xml)
- docs sitemap should include all pages

kamaldeepcheema-okta:
> @francoislehoux-okta @kamaldeepcheema-okta if the goal is to take over the sitemap, I would recommend:
> 
> * creating a static file sitemap_index.xml with two links (docs sitemap - the name doesn't matter) + blog sitemap - sitemap.xml)
> * docs sitemap should include all pages

I've updated the PR now to include a static sitemap_index.xml file that lists both docs-sitemap.xml (which includes all the files) and sitemap.xml (which is the blog's sitemap).

@francoislehoux-okta @hamzaessaoui-okta

shawyu-okta:
Do we need to update the generation of `robots.txt` so that it points to the `sitemap_index.xml` now?

francoislehoux-okta (approved):

kamaldeepcheema-okta:
> Do we need to update the generation of `robots.txt` so that it points to the `sitemap_index.xml` now?

Yeah that's a good idea, it would make more sense.

hamzaessaoui-okta:
@kamaldeepcheema-okta yes please, it would be great if both changes go live together

shawyu-okta (approved):

kamaldeepcheema-okta:
> @kamaldeepcheema-okta yes please, it would be great if both changes go live together

Yep I just updated the PR for this change

hamzaessaoui-okta:
Awesome! could you please ping me when this is live? I'll update Coveo

PR #1883, opened 2021-02-04
added table row

christinechoy-okta:
## Description:
- **What's changed?** Add to rate limit authentication table
- **Is this PR related to a Monolith release?** Yes Release 2021.03.0

PR #1923, opened 2021-02-19
Revert "added table row"

susanharper-okta:
Reverts okta/okta-developer-docs#1883

This is for the March release. So, it needs to be reverted.


susanharper-okta (commented):
LGTM

christinechoy-okta (approved):

PR #1940, opened 2021-02-25
Updating the deployment checklist for email pain point

albertchen-okta:
## Description:
Added a bullet point calling out bulk user migration as a step for the deployment checklist, and added a blurb to ask user to check their email notification settings.

### Resolves:

This was a request from product leadership (Grace W and Arnab):
1. Today, MFA reset emails are set to on by default. 
2. Recently, major customers (Cisco, Pinterest) have run into issues with MFA reset emails being sent to millions of customers as part of an internal migration process and are asking that these be turned off by default. These are generally CIAM use cases. 
3. However, from a workforce use case perspective, there is benefit and value to having these turned on by default as this helps drive education and adoption and 77% of our customers leave this on.

The decision is we will turn off email notifications by default, but it's important regardless of the default on or off, that we should update the docs in case a customer developer does not know that step is worth checking prior to bulk migrating users.

### DIFF
--- a/packages/@okta/vuepress-site/docs/guides/production-deployment/deployment-checklist/index.md
+++ b/packages/@okta/vuepress-site/docs/guides/production-deployment/deployment-checklist/index.md
@@ -28,6 +28,7 @@ Ensure your application fits your brand:
 
 ### User authorization and registration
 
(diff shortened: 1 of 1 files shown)


albertchen-okta:
For reference, this is the Admin UI
Screen Shot 2021-02-24 at 5 26 16 PM

jakubvul-okta (approved):

PR #1954, opened 2021-03-01
Add CORS Support Icon for /token

Tennyx:
/token is CORS enabled as of 2019 for flows like PKCE. Unless there is something special/different about this endpoint as it relates to CORS, I think we should add the icon for consistency's sake.

## Description:
- **What's changed?** 

### Resolves:


shantanusardal-okta:
The token endpoint is not CORS enabled in the traditional sense. We add CORS headers only under certain circumstances - https://okta.slack.com/archives/C7KRAC4SF/p1614804618233600?thread_ts=1614804568.233500&cid=C7KRAC4SF

Unless we add more details on how this endpoint supports CORS, we should not add the <SupportsCors /> icon.

PR #1975, opened 2021-03-09
Update accesstoken.md

tommycwu:
The ValidateAudience parameter is missing.  There is a comment regarding the audience, but it says you should validate it but the code sample doesn't have anything in it's place and it's a runtime error that isn't very helpful.  I had a customer read that and assume that value was optional therefore didn't put it in, kept getting errors and even PS couldn't figure it out... it got to me and I figured it out but the customer had wasted a week by then and was frustrated.  I put in the value = false, but if you want to have the code sample set that to true with the additional parameter filled out that's fine too.  e.g..

ValidateAudience = true,
ValidAudience = "api://default",

## Description:
- **What's changed?** 

### Resolves:


andriizhegurov-okta (approved):
I think this change makes sense, default value for `ValidateAudience` is true.

PR #2104, opened 2021-05-03
Update index.md

gabrielsroka:
`RADIUS_FOR_GENERIC_APP` was not listed in Sign-On Modes. can we add relevant info and any other Modes that are missing (eg `RADIUS_FOR_CISCO_ASA`, `RADIUS_FOR_CITRIX_NETSCALER`, etc, etc)?


jakubvul-okta:
After consulting with team, we are not going to make this change right now.

PR #2106, opened 2021-05-03
/keys best practice

Tennyx:
I added a Best Practice section to /keys as discussed 

Two things:

1. I added in the detailed best-practice advice laid out in the internal Jira comments by Jon Kraatz here:

https://oktainc.atlassian.net/browse/OKTA-381855

2. I also moved the below note from higher up in the docs to this Best Practices section:

> **Note:** Okta strongly recommends retrieving keys dynamically with the JWKS published in the discovery document.<p>
Okta also recommends caching or persisting these keys to improve performance. If you cache signing keys and automatic key rotation is enabled, be aware that verification fails when Okta rotates the keys automatically. Clients that cache keys should periodically check the JWKS for updated signing keys.

Moving that down feels...wrong in a way because it's less "in-your-face" - but the long term goal/idea would be to establish this section in as many parts of our docs as we can. Once customers become familiar with it, then they will go looking for it in our docs. And customers who are not aware will still get there eventually if they're being thorough.

I think this can be useful to establish more documented best practices, as well as allow some separation in our docs between technical info and implementation advice.

## Description:
- **What's changed?** 

### Resolves:


susanharper-okta (commented):
couple things

susanharper-okta (approved):
LGTM 🚢

PR #2181, opened 2021-06-01
Add robots meta tag

volodymyrmakarchuk-okta:
## Description:
- **What's changed?** 

### Resolves:

* [OKTA-253160](https://oktainc.atlassian.net/browse/OKTA-253160)

### DIFF
--- a/packages/@okta/vuepress-site/.vuepress/config.js
+++ b/packages/@okta/vuepress-site/.vuepress/config.js
@@ -396,7 +396,7 @@ module.exports = ctx => ({
         if (!page.frontmatter['meta']) {
           page.frontmatter['meta'] = []
         }
-        page.frontmatter['meta'].push({ name: 'robots', content:'noindex'});
+        page.frontmatter['meta'].push({ name: 'robots', content:'noindex,nofollow'});
       });
     }
   },
--- a/packages/@okta/vuepress-site/.vuepress/config.js
+++ b/packages/@okta/vuepress-site/.vuepress/config.js
@@ -389,7 +389,7 @@ module.exports = ctx => ({
     }
   },
   async ready() {
-    if (process.env.ENV !== 'prod') {
+    if (process.env.DEPLOY_ENV !== 'prod') {
       ctx.pages.forEach((page) => {
         // We adding meta tag `robots` for all non-prod versions of the site
         // to be able to exclude test envs from browser search.
--- a/packages/@okta/vuepress-site/.vuepress/config.js
+++ b/packages/@okta/vuepress-site/.vuepress/config.js
@@ -11,7 +11,7 @@ const outputDir = Path.resolve(__dirname, '../dist/');
 
 const WIDGET_VERSION = findLatestWidgetVersion(signInWidgetMajorVersion);
 
-module.exports = {
+module.exports = ctx => ({
   dest: 'dist',
   theme: "@okta/vuepress-theme-prose",
   /**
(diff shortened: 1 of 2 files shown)


packages/@okta/vuepress-site/.vuepress/config.js
```diff
@@ -389,4 +389,14 @@ module.exports = {
       }
     }
   },
-}
+  async ready() {
```
shawyu-okta, line 392: This is a really good find -- I didn't know we had the ability to iterate on `ctx.pages` and then update the content per-page like this. This may be helpful in the future for page-specific header content.

packages/@okta/vuepress-site/.vuepress/config.js
```diff
@@ -389,4 +389,14 @@ module.exports = {
...
+    if (!ctx.isProd) {
```
shawyu-okta, line 393: As mentioned in Slack, this won't work since both the Trex and Prod environment deploys use the "production" webpack build, which will result in `ctx.isProd = true`

shawyu-okta (changes requested):
marked one issue we discussed on slack

scripts/publish.sh
```diff
@@ -5,12 +5,16 @@ cd ${OKTA_HOME}/${REPO}/packages/@okta/vuepress-site
...
 export REGISTRY_REPO="npm-topic"
 export REGISTRY="${ARTIFACTORY_URL}/api/npm/${REGISTRY_REPO}"
-export TEST_ENV="true"
+export PROD_ENV=""
 
 declare -A branch_environment_map
 branch_environment_map[master]=vuepress-site-prod
 branch_environment_map[staging]=vuepress-site-preprod
 
+if [[ $BRANCH == "master" ]]; then
```
shawyu-okta, line 14: Leave a comment here indicating why we use the `PROD_ENV` variable

packages/@okta/vuepress-site/.vuepress/config.js
```diff
@@ -389,4 +388,14 @@ module.exports = {
...
     }
   },
-}
+  async ready() {
+    if (!process.env.PROD_ENV) {
+      ctx.pages.forEach((page) => {
+        if (!page.frontmatter['meta']) {
+          page.frontmatter['meta'] = []
+        }
+        page.frontmatter['meta'].push({ name: 'robots', content:'noindex,nofollow'});
```
shawyu-okta, line 397: Add a comment here to mention why we are adding the meta tag

shawyu-okta (approved):

packages/@okta/vuepress-site/.vuepress/config.js
volodymyrmakarchuk-okta: done

robertnicolo-okta: I can understand adding **noindex** to block the search engines from indexing the page but why are we adding **nofollow**?  **nofollow** is used when the search engine crawls through other links on your page.

scripts/publish.sh
robertnicolo-okta: Instead of PROD_ENV, wouldn't it be better to have a ENV variable with values of "test" and "prod".  This will lend better support for additional environments or if an environment name changes in the future.

robertnicolo-okta (changes requested):

packages/@okta/vuepress-site/.vuepress/config.js
shawyu-okta: Since this is only for the preview trex site, we might have links to pages that are also not public yet, so it's reasonable to say `nofollow` as well here.

scripts/publish.sh
shawyu-okta: I suppose we could call this `DEPLOY_ENV` and set it to `"production"` to leave flexibility for the future

shawyu-okta (approved):
one last comment but otherwise lgtm

robertnicolo-okta (approved):
lgtm

PR #2185, opened 2021-06-02
Fix Inline Password Hook link

jvandyke:
## Description:
- **What's changed?** 
  - I noticed that this link was incorrect while reading. I am assuming that the link was meant to point to the feature reference and not the guide to migrate users using the feature, but correct me if I'm mistaken.
  - Not that I know of.

### Resolves:


brianduffield-okta (approved):
LGTM. Good catch.

PR #2188, opened 2021-06-03
clarify that the UTC Epoch time is in seconds

ryber:
## Description:
In many systems (like Unix, Java, etc) Epoch timestamps are in Millies, this one is in seconds but the doc doesn't say that explicitly (if you look at the example it's clearly seconds). This just clarifies that.


susanharper-okta (approved):
one minor update. ship it!

PR #2205, opened 2021-06-08
Minor improvements & nits

tmarshall:
## Description:

While implementing an auth flow I came across a few issues in the docs. This PR is a mix of fixes & nits.

- `/userinfo` is listed as a `GET` but the example given is a `POST`—I've updated it to `GET`
- There was a visible trailing `)` on one page
- The `/token` endpoint has a hard requirement for a `client_id`, but the page does not list that as a param.
- Minor syntax cleanup


susanharper-okta (commented):
one suggested change

susanharper-okta (approved):
LGTM

PR #2208, opened 2021-06-09
Moving note snippet above ordered list, to correct numerics

tmarshall:
## Description:

Fixes an issue where a snippet split ordered list items, causing the numerics to reset.

Screenshot_6_9_21__10_50_AM

By moving the snippet above the list it will allow the numerics to show the expected values.


vanngo-okta (approved):
LGTM!

PR #2214, opened 2021-06-11
Renamed headings and fixed linking to avoid page autoscroll on top

AndriiDovhopolov-okta:
## Description:
- Name of  all `/` links in "Create Routes" section of pages which include `_sign-in_widget/` info is changed to `/ - index page`
image
image
- Links in upper part of pages where adjusted to new routes

### Resolves:
* [OKTA-397842](https://oktainc.atlassian.net/browse/OKTA-397842)


shawyu-okta (commented):
Can you revert the automatic format/style changes in this PR? It's very difficult to review what is the actual fix versus the numerous autolinter changes. Perhaps we can run an autolint PR separately so this is not an issue in the future.

shawyu-okta (approved):
Let's get @okta/developerdocs for a review before merge

PR #2249, opened 2021-06-22
Edits for oie-embedded-common-org-setup

susanharper-okta:
## Description:
- **What's changed?** Edits for oie-embedded-common-org-setup 

### DIFF
--- a/packages/@okta/vuepress-site/docs/guides/oie-embedded-common-org-setup/index.md
+++ b/packages/@okta/vuepress-site/docs/guides/oie-embedded-common-org-setup/index.md
@@ -1,6 +1,6 @@
 ---
-title: Create and setup your Okta org
-excerpt: Create and setup your Okta org
+title: Create and set up your Okta org
+excerpt: Create and set up your Okta org
 layout: Guides
 sections:
  - main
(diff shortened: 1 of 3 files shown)


robertnicolo-okta (approved):
My feedback from Susan's comments:

1a: Susan comment 1 for: "I don't think 1a, 1b, 1c are necessary....these could be separate headers and not be confusing)"
1b: Robert response: Will remove the letters in the steps in a followup PR."" 


2a: Susan comment 2 for: "(I don't see a dialog box after clicking edit - I see an entire page of fields and these are a part of that page)" in the "Step 1b: Set up the password authenticator for authentication" section.
2b: Robert response:  I don't see it either in the latest org I created. This is a big issue. For now let's delete this section.  I'll be going over the entire flow again today and and readd a section if it's needed for the password authenticator. 

3a Susan comment 3 for: "this section does not appear for me" in the "Add the phone authenticator" section.
3b: Robert Response:  Let's keep this one as we discussed today, since its in the OK12 orgs.


4a: Susan comment: " ??In the **Switch to Live Mode?** dialog box, click **Switch Mode**.?? (I don't get this dialog box in Facebook Dev)":  
4b: Robert Response: I just tested and I see the dialog window.  I wonder whether the facebook for developers is versioned like Okta orgs? Want to keep it for now? I like to get these changes into the buildathon tomorrow but we can talk more about this particular discrepancy tomorrow in a zoom sync up.

5a Susan comment: "??Keep **Idp Usage** set to the **SSO Only** default.?? (I do not see this field)""
5b: Robert Respnse:  I think this is an org version issue.  Let's keep it and I can show you tommorrow the field in the newer org if you like.

6a Susan comment: "Keep the **Scopes** values set to the default: **public_profile** and **default**.??? (I see **email** not default?)"
6b Robert Resposne: I don't see "default" either in the newest org I created. Lets change it to email.

PR #2273, opened 2021-06-25
Change pricing link

volodymyrmakarchuk-okta:
## Description:
- **What's changed?** 

### Resolves:

* [OKTA-404731](https://oktainc.atlassian.net/browse/OKTA-404731)

### DIFF
--- a/tests/selenium/framework/page-objects/MainPage.js
+++ b/tests/selenium/framework/page-objects/MainPage.js
@@ -7,7 +7,6 @@ const signUpButtonSelector = '.sign-up--button';
 const signupPricingSelector = '.pricing-link'
 const signInLinkXpath = "//*[text()='Sign in to Okta']";
 const oktaMainPageLinkXpath = "//*[text()='Okta.com']";
-const pricingLinkXpath = "//*[text()='Pricing']";
 const communityLinkXpath = "//*[text()='Community']";
 
 class MainPage extends BasePage {
--- a/tests/selenium/framework/page-objects/MainPage.js
+++ b/tests/selenium/framework/page-objects/MainPage.js
@@ -4,9 +4,9 @@ const BasePage = require('./BasePage');
 
 const headerSelector = '.page-header';
 const signUpButtonSelector = '.sign-up--button';
-const signupPricingSelector = '.pricing-link'
 const signInLinkXpath = "//*[text()='Sign in to Okta']";
 const oktaMainPageLinkXpath = "//*[text()='Okta.com']";
+const pricingLinkXpath = "//*[text()='Pricing']";
 const communityLinkXpath = "//*[text()='Community']";
 
 class MainPage extends BasePage {
@@ -57,7 +57,7 @@ class MainPage extends BasePage {
   }
 
   getPricingLinkElement() {
-    return element(by.css(signupPricingSelector));
+    return element(by.xpath(pricingLinkXpath));
   }
 
   getCommunityLinkElement() {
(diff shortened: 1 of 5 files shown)


tests/selenium/spec/header-spec.js
```diff
@@ -28,16 +29,21 @@ describe('header sanity check', () => {
 
 
   describe('validate sign up page elements', () => {
+    const pricingURL = 'https://www.okta.com/pricing/#customer-identity-products'
```
shawyu-okta, line 32: small nit: we can put this inside the test block since it is only used there

tests/selenium/framework/page-objects/MainPage.js
```diff
@@ -57,7 +58,7 @@ class MainPage extends BasePage {
   }
 
   getPricingLinkElement() {
-    return element(by.xpath(pricingLinkXpath));
+    return element(by.css(signupPricingSelector));
```
shawyu-okta, line 61: Do we have to use a css selector for this? Why won't the XPath selector work?

volodymyrmakarchuk-okta: yes. so selenium test will open a new page with okta.com site and it pricing page also includes `Pricing` inside, so in console, In such case, I see the warning `more than one element found for locator By(xpath, //*[text()='Pricing']) - the first result will be used`. So that's why I decide to use more unique selector

shawyu-okta: Is there a way we can still keep using xpath but use a more specific path expression? I'd like to avoid adding special logic in the footer rendering just to insert the CSS class for a selenium test.

shawyu-okta (approved):
One comment -- try to see if you can find a better way and then lgtm

tests/selenium/spec/header-spec.js
```diff
@@ -34,10 +35,16 @@ describe('header sanity check', () => {
...
     }));
 
-    it('pricing link click takes user to pricing page', util.itHelper(async () => {
+    it('pricing link click takes user to pricing okta.com page from a new tab', util.itHelper(async () => {
+      const pricingURL = 'https://www.okta.com/pricing/#customer-identity-products';
+
       await mainPage.getPricingLinkElement().click();
-      expect(await mainPage.getCurrentURL())
-        .to.contain("/pricing/");
+      await browser.sleep(10000);
```
robertnicolo-okta, line 42: How did we choose 10 seconds?  Wouldn't the wait method be a better solution here. Network or performance issues could make the wait more than 10 seconds. Also, we are adding another 10 seconds to be build process (>45 mins) when it could be less time depending on network, cpu, etc.  For more info on the wait method see:

https://riptutorial.com/protractor/example/26620/browser-sleep---vs-browser-wait--
https://stackoverflow.com/questions/45584840/what-is-the-difference-between-browser-sleep-and-browser-wait-methods

robertnicolo-okta (changes requested):

tests/selenium/spec/header-spec.js
volodymyrmakarchuk-okta: I've made some edits, `sleep` method doesn't require here

shawyu-okta (approved):
@robertnicolo-okta can you please re-review?

robertnicolo-okta (approved):
I have lgtm'd the PR but let's keep a lookout to see whether the errors this change is supposed to fix still occurs.

PR #2293, opened 2021-07-06
Update index.md

gabrielsroka:
i don't think this works anymore.

maybe
click your name in the upper-right corner of the ADMIN console
hover over the name and click Copy


susanharper-okta:
This is still very much the case in the Global Header of the Admin Console:

Screen Shot 2021-07-26 at 5 22 15 PM

PR #2339, opened 2021-07-16
feat: single stack selectors for all snippets on the page

oleksandrpravosudko-okta:
## Description:
- **What's changed?** Stack Selector decoupled 
### Resolves:

* [OKTA-403629](https://oktainc.atlassian.net/browse/OKTA-403629)


AndriiDovhopolov-okta:
Netlify: https://60feb7454a44d0247110df71--reverent-murdock-829d24.netlify.app

robertnicolo-okta (changes requested):
See following items:

1. Use case overview page is showing no code snippets defined.  See screenshot below:
image

2. Double tech selector dropdowns noticed on use case pages
image

3. Nit: Dropdown and dropdown text is not aligned with other headings and text on the page:
image

4. The dropdown text interferes with other page text when scrolling. Add a white box background, perhaps?
image

5. Is there a behavior we are trying to achieve with the dropdowns? For example, hide one and not the other on scrolling.
image

robertnicolo-okta (changes requested):
Another item:

1. On smaller screens and mobile client displays, the tech stack selector is on top per the figma mockups: 

**Current behavior:**
image

**Mockups:**
image

robertnicolo-okta (changes requested):
Here is the figma mockup link for reference:

https://www.figma.com/file/0R4eMiDio46zWRaBxl28bW/%5BOIE%5D-Page-level-stack-selector?node-id=1%3A16180

chrisdmills-okta (changes requested):
My 0.10 cents here — I agree with @robertnicolo-okta 's comments here for the most part, and don't have much else to add. But for the record:

1. Desktop/widescreen layout.

I think the language selector should always appear in the right-hand sidebar, and never at the top of the page content. I've always been a bit uneasy about the great big drop-down being stuck at the top of the page content — it looks unwieldy and interferes with the page content too much. At the end of the day, it's really metadata, or an admin setting, not part of the page content.

I've noticed that it interferes with the right-hand nav menu when one exists and the nav menu starts to scroll:



Screenshot 2021-07-27 at 09 26 54


(for reference, this is taken from https://60feb7454a44d0247110df71--reverent-murdock-829d24.netlify.app/docs/guides/sign-into-spa/angular/before-you-begin/#refresh-tokens-and-spas)

Robert is right that we should give the language selector a white background; I also think perhaps we should position the nav menu's scrolling region down the page a bit, so that it never reaches the language selector, and also look to make the scrolling behavior a bit smoother. It can be a bit jarring currently. 

In the page referenced above, the menu starts to jarringly scroll upwards when you start to scroll through the "Refresh tokens and SPAs" section. IMO, the nav menu styling should update to show what page section you are currently inside — that makes sense — but the menu shouldn't scroll at all if there are only a small number of menu items so they can all comfortably be seen. The menu only needs to scroll at the point where there are enough menu items that it is possible that the currently-viewed page section cannot been seen in the menu. In such a case, the menu should smoothly scroll to show the relevant menu item.

---

The "No code snippets defined" message on pages where there are no language-specific instructions is a bit concerning to me — I'm pretty sure users will consider this to be a global setting, not a per-page setting (I personally do), so suddenly seeing that message might make them think the page/article series has a bug. It would be much better to just carry on showing the "Instructions for" drop down with the last known language setting selected.

On some pages we have a choice of server-side languages/platforms, on some pages we have a choice of client-side JS frameworks, and on some pages we have a choice of social IdPs (are there more?) I'm assuming we store each of these settings in separate cookies or local storage items, and then for each page we get to, we can figure out which one is more appropriate to show? E.g. "This series of articles is about Java code, so on pages that don't have code snippets explicitly defined, show "Instructions for Java"". Do we have any article series where we have more than one type of choice on different pages of the series? That's where this would start to get a bit hairy, although I guess we could start to show multiple selectors in such an eventuality.

One more point to make here — eventually we are going to start amalgamating some of the multiple page article series into single articles, as they feel a bit "bitty" and annoying the navigate through. In such cases, we may end up with multiple stack selector sections on each page, so the global setting thing starts to make even more sense.

---

2. mobile/narrow screen layout.

When I checked the site out on mobile/narrow screen view, I didn't see the language selector at all (maybe it was because I was using Firefox Responsive Design Mode rather than a real device), but it looks to me like the Figma mockup makes sense — the language selector is metadata, so should be outside the page content, up top, whereas the on this page dropdown should  be closer to the content.

I did wonder whether it would make sense to use `position: sticky` to anchor the on this page dropdown to the top of the page when you start scrolling, or whether it wouldn't be worth it in terms of number of people using it versus loss of page real estate...

AndriiDovhopolov-okta:
@robertnicolo-okta @chrisdmills-okta 

here is updated netlify with bugfixes. https://6103f5fce8288d5ff391a4db--reverent-murdock-829d24.netlify.app

@chrisdmills-okta also, stack selector  is working not the way you think it is working (code-wise, I mean) - there are no any direct data flows (cookies, local storage, services, etc). I can describe what is it doing next week if you are interested in a deep dive. It's kinda tricky.

chrisdmills-okta:
> @chrisdmills-okta also, stack selector is working not the way you think it is working (code-wise, I mean) - there are no any direct data flows (cookies, local storage, services, etc). I can describe what is it doing next week if you are interested in a deep dive. It's kinda tricky.

OK, thanks for the heads-up. Yeah, I might be interested in hearing a bit more about that.

chrisdmills-okta (changes requested):
Thanks for the update @AndriiDovhopolov-okta ! This does look a lot better now; the placement issue are sorted, and the weird interference with the right-hand menu.

I did still have a nitpick about the dropdown menu itself:

Screenshot 2021-07-30 at 15 15 19
 
The icon needs better alignment with the text and the arrow, and it also looks like there should be a bit more padding on the right-hand side.

alexdahl-okta (changes requested):
This seems to have diverged substantially from the design, please make sure to get review from Christine Asprea.

chrisdmills-okta (changes requested):
Hi @AndriiDovhopolov-okta ! I've had a look at the latest changes, and feel that this still needs some more work.

Looking at the default state of the stack selector:

Screenshot 2021-09-02 at 10 25 24

* There isn't enough margin between the stack selector and the content below it.
* In the stack selector itself, the "open" arrow and the platform/language icon don't look vertically centered.
* I also noticed that when you hover over the stack selector, the mouse pointer adopts the text cursor/caret icon. Surely this should be the hand/pointer icon, the same as you get when you hover over links, to indicate that something is actionable/clickable here?

Now if we look at the open stack selector:

Screenshot 2021-09-02 at 10 26 03

* The different choices seem cramped/too close together. They could use some more margin between each other, and the bottom of the container
* If you look at the blue hover box, you can see that the platform icon and language name are not vertically centered in it. They should be centered in their respective space.

Finally, the mobile view is a bit of a mess:

Screenshot 2021-09-02 at 10 26 21

* The color/position/style of the stack selector is significantly different to that of the "On this page" drop down menu. Even the capitalization of the two drop down labels is different.
* The stack selector, h1, and "On this page" menu are all fixed position, therefore on a small screen you can barely see the content. This could really do with a significant redesign, although I think the quick fix for now would be to have the heading and stack selector not fixed on small screen layout.

robertnicolo-okta (changes requested):
My feedback is described in the following [video](https://okta.box.com/s/zrfltpw2a4qyehkx3s6rx1eecaljs3qt).

**Summary of issues:**
1. No border around the dropdown
2. Dead selection space between items in the dropdown
3. Dropdown styling different from other dropdowns (same issued identified by Chris).

chrisdmills-okta:
Stack selector work now done in a different PR, so closing this out.

PR #2401, opened 2021-08-03
Returns build configuration to use `medium` instances

shawyu-okta:
## Description:
- **What's changed?** 
* Returns us to `medium` instances (free/included with our Travis plan)
* Updates build to use thread pool-like behavior (we actually discard and recreate threads) with a max number of pages per thread to limit memory usage
* We may want to use a real thread pool implementation library at some point to be even more efficient in the future

### Resolves:

* [OKTA-417069](https://oktainc.atlassian.net/browse/OKTA-417069)

### DIFF
--- a/packages/@okta/vuepress-site/.vuepress/scripts/updateBuildScript.js
+++ b/packages/@okta/vuepress-site/.vuepress/scripts/updateBuildScript.js
@@ -5,7 +5,7 @@ const webpack = require('webpack')
 const readline = require('readline')
 const { Worker } = require('worker_threads')
 
-const workerThreads = 8
+const workerThreads = 6
 
 const {
   chalk,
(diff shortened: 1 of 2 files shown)


packages/@okta/vuepress-site/.vuepress/scripts/updateBuildScript.js
```diff
@@ -19,15 +17,23 @@ const createClientConfig = require('../webpack/createClientConfig')
 const createServerConfig = require('../webpack/createServerConfig')
 const { applyUserWebpackConfig } = require('../util/index')
 
+const MAX_WORKER_THREADS = 6;
+const PAGES_PER_THREAD = 100;
```
shawyu-okta, line 21: We can tune this more for better performance but I'm seeing consistent passes right now so I'm leaving it this way. It appears 8 threads is too many for the VM memory.

AndriiDovhopolov-okta (approved):
Looks reasonable, do not have objections.

PR #2511, opened 2021-08-31
Android Self Registration Updates

susanharper-okta:
## Description:
- **What's changed?** Android Self Registration Updates

### Resolves:

* [OKTA-423086](https://oktainc.atlassian.net/browse/OKTA-423086)


vanngo-okta (approved):
I like how you broke down step 8. It's a much better way to represent the options (we should update the Java code in a similar fashion)
I have a few small nits with the diagrams:
* step 6 uses "email factor", but step 8 uses "phone authenticator" (this problem was probably in the image from before, just wondering which term should we use?)
* step 3 uses "enrollment factors" and step 4 uses "authenticator to enroll" (again, this is probably from previous diagram)
* 8 (option 2.2) ?
* step 9 should be part of step 8 option 2 ?

susanharper-okta:
> I like how you broke down step 8. It's a much better way to represent the options (we should update the Java code in a similar fashion)
> I have a few small nits with the diagrams:
> 
> * step 6 uses "email factor", but step 8 uses "phone authenticator" (this problem was probably in the image from before, just wondering which term should we use?)
> * step 3 uses "enrollment factors" and step 4 uses "authenticator to enroll" (again, this is probably from previous diagram)
> * 8 (option 2.2) ?
> * step 9 should be part of step 8 option 2 ?

It's authenticator - just verified with InfoDev/Graham. The term factor is still used, but not when referring to say SMS, phone, password, and so on. Factor is used in situations such as: knowledge-based factor, possession factor, etc

vanngo-okta (approved):
LGTM!

PR #2536, opened 2021-09-03
Updated image wth John Doe

robertnicolo-okta:
## Description:
Replaced user with John Doe in the Golang signed in page image.

### Resolves:

* [OKTA-424450](https://oktainc.atlassian.net/browse/OKTA-424450)


chrisdmills-okta (approved):
The updated image looks good to me, but is there any particular guideline that we use for not showing sensitive data? The original JIRA ticket mentions blurring out the data, although I think just using an example value looks cleaner.

@susanharper-okta probably knows the answer to this ;-)

PR #2546, opened 2021-09-08
Removing the alternate flow page and moving content to self-reg

susanharper-okta:
## Description:
- **What's changed?** Removing the alternate flow page and moving content to self-reg
- **Is this PR related to a Monolith release?** NO


robertnicolo-okta (approved):
lgtm

PR #2616, opened 2021-10-07
Explain that the EU and AP continent codes do not include all the countries in those continents.

mikepatnode-okta:
## Description:
- **Add a note explaining what the EU and AP continent codes are used for**

### Resolves:

* [OKTA-431277](https://oktainc.atlassian.net/browse/OKTA-431277)

### DIFF
--- a/packages/@okta/vuepress-site/docs/reference/api/zones/index.md
+++ b/packages/@okta/vuepress-site/docs/reference/api/zones/index.md
@@ -115,6 +115,8 @@ Each Location object specifies an [ISO-3166-1](https://en.wikipedia.org/wiki/ISO
 | country        | Format of the value: length 2 ISO-3166-1 country code                | String      | Yes       |
 | region       | Format of the value: `countryCode`-`regionCode`, or `null` if empty  | String      | No       |
 
+> **Note**: Continent codes are not intended to be used as `country` or `region` definitions. To include all of the countries in a continent, each country needs to be specified individually in a Location definition.  If a continent code is used, then traffic from a designated country region within the continent might not be considered. For example, if `EU` or `AP` is used instead of a country code, then the Location definition treats Europe or Asia/Pacific as generic codes for undesignated regions rather than inclusive of the countries they contain.
+
 #### Location object example (with region)
 
 ```json


rpamidimarri-okta (approved):

packages/@okta/vuepress-site/docs/reference/api/zones/index.md
```diff
@@ -115,6 +115,8 @@ Each Location object specifies an [ISO-3166-1](https://en.wikipedia.org/wiki/ISO
 | country        | Format of the value: length 2 ISO-3166-1 country code                | String      | Yes       |
 | region       | Format of the value: `countryCode`-`regionCode`, or `null` if empty  | String      | No       |
 
+> **Note**: Continent codes are not intended to be used as region definitions. The EU and AP codes are only used if a specific country code has not been set. To include all of the countries in Europe or in Asia/Pacific, each country needs to be specified individually.  Used alone, Europe and Asia/Pacific are treated as generic codes for undesignated regions rather than inclusive of the countries they contain.
```
vanngo-okta, line 118: Can you provide an example of a continent code in lieu of country code? It is unclear what the format should be?
"country": null,
"region": "-AP" ?
"region": "AP" ?
Also, is `regionCode` == continent code?

vanngo-okta (commented):
Sorry, I have a question regarding what to expect in the location object if you only have the continent code specified. Does the Location object table need to be updated?

packages/@okta/vuepress-site/docs/reference/api/zones/index.md
vanngo-okta: I'm reading the Jira ticket and for the API docs, I think we may need to clarify what type of data is allowed in the Location object. Suggestion to update the Location object property table:

| Field Name  | Description                                                | Data Type   | Required |
| :---------- | :--------------------------------------------------------- | :---------- | :------- |
| country        | The country included in the Zone perimeter. Specify the [ISO 3166-1 alpha-2](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2) country code.               | String      | Yes       |
| region       | The country subdivision included in the Zone perimeter. Specify the [ISO-3166-2](https://en.wikipedia.org/wiki/ISO_3166-2) region code, in the format: `countryCode`-`regionCode`. This field is optional: if you specify the subdivision region code, then only this region is considered in the Zone perimeter; if you specify `null` in this field, then the entire country is considered in the Zone perimeter.   | String      | No       |

> **Note:** You cannot specify a continent code in the `country` or `region` fields, for example `EU` for Europe. If you want to include all of the countries in a continent, you need to specify a location for each country. If you specify an unknown country or region code, then the Zone perimeter does not include any location.

packages/@okta/vuepress-site/docs/reference/api/zones/index.md
```diff
@@ -115,6 +115,8 @@ Each Location object specifies an [ISO-3166-1](https://en.wikipedia.org/wiki/ISO
 | country        | Format of the value: length 2 ISO-3166-1 country code                | String      | Yes       |
 | region       | Format of the value: `countryCode`-`regionCode`, or `null` if empty  | String      | No       |
 
+> **Note**: Continent codes are not intended to be used as country or region definitions. The EU and AP codes are only used if a specific country code has not been set. To include all of the countries in Europe or in Asia/Pacific, each country needs to be specified individually.  Used alone, Europe and Asia/Pacific are treated as generic codes for undesignated regions rather than inclusive of the countries they contain.
```
vanngo-okta, line 118:
```suggestion
> **Note**: Continent codes are not intended to be used as `country` or `region` definitions. To include all of the countries in a continent, each country needs to be specified individually in a Location definition.  If a continent code is used, then traffic from a designated country region within the continent might not be considered. For example, if `EU` or `AP` is used instead of a country code, then the Location definition treats Europe or Asia/Pacific as generic codes for undesignated regions rather than inclusive of the countries they contain.
```

vanngo-okta (changes requested):
I've made another suggestion. Can you please take a look at see if it makes sense? I'm trying to explain it in a generic way so that someone who is configuring this Zone would know what would happen if they use a continent code. Should we mention that IPs with no country or region designation, but are considered part of the continent by the geolocation provider is considered?

PR #2629, opened 2021-10-13
Use semver comparator to sort widget versions correctly

shawyu-okta:
## Description:
When getting the latest widget version to display in docs and live widget, we sorted the list of versions from NPM, but were sorting semver strings. When the version number has a two digit value in any place, the sorting will become out of order numerically, even though lexically the sort is accurate. (e.g. `"5.1.10"` is seen as _less than_ `"5.1.9"` from a lexical perspective, but numerically we know that it should be _greater_). Using `semver` comparators in the sort function solves this.


albertchen-okta (approved):
netlify looks good, thanks

PR #2789, opened 2021-11-16
updates to Angular quick start

alisaduncan:
## Description:
- **What's changed?** Update Angular Quickstart to use correct Okta SDKs so a developer successfully creates an authenticated app. I tested using Angular v13. I think the only place where prior versions of Angular might have to make a change is where the access token is used via the HTTP call due to a change in RxJS v7. 
- **Is this PR related to a Monolith release?** I don't think so?

### Resolves:

* No Jira ticket created (yet).


shuowu-okta (approved):

brentschaus-okta (approved):
LGTM

PR #2805, opened 2021-11-17
OKTA-447602 fixes issues with TBS Napa docs

brentschaus-okta:
## Description:
- **What's changed?** Small fixes to my Napa docs.
- **Is this PR related to a Monolith release?** Nein.

### Resolves:

* [OKTA-447602](https://oktainc.atlassian.net/browse/OKTA-447602)

### DIFF
--- a/packages/@okta/vuepress-site/docs/guides/oie-upgrade-sign-in-widget-deprecated-methods/main/index.md
+++ b/packages/@okta/vuepress-site/docs/guides/oie-upgrade-sign-in-widget-deprecated-methods/main/index.md
@@ -5,7 +5,7 @@ title: Deprecated JavaScript methods in the Sign-In Widget
 <ApiLifecycle access="ie" /><br>
 <ApiLifecycle access="Limited GA" /><br>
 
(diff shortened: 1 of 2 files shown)


brianduffield-okta (commented):
Line 8 "Javascript" > "JavaScript" (In deprecated doc)

brianduffield-okta (commented):
Line 30 and 31. Both Notes are rendering as one on my local. Might need a carriage return or other formating change. ( In deprecated doc)

brianduffield-okta (commented):
Two references to https://github.com/okta/okta-auth-js, which are different (Auth.js and Auth.js API reference.) We should probably standardize on terminology. I used full SDK name in the mapping doc: Okta Auth JavaScript SDK, but internally and probably externally it's known as Auth JS. At least, they should probably the same in the doc.

brianduffield-okta (commented):
Just as an overall comment, my SME, Shuo, asked me to specifically leave out the braces on all methods documented, i.e., to not use `setCookieAndRedirect()` but rather simply `setCookieAndRediret`. I believe the rationale is that empty braces imply a specific call (no parameters). Recommend you confirm with your SME. (In deprecated doc)

PR #2826, opened 2021-11-19
remove insecure link from dev.okta

chrisdmills-okta:
## Description:
- **What's changed?** We were informed that the base64encode.org/base64decode.org site that we are linking to is insecure; this PR removes mention of it. I've restructured the test so that (imo) we don't really need it any more, and added it to one of the other stack selectors that looks like it also needs those instructions.


packages/@okta/vuepress-site/docs/guides/implement-grant-type/main/saml2assert/use-flow.md
```diff
@@ -38,3 +38,43 @@ Note the parameters that are being passed:
     "id_token": "eyJraWQ.....h7BYbgCzQ"
 }
 ```
+
```
vanngo-okta, line 41: I don't think you need to add this section for SAML. The SAML assertion is not the same thing as the client ID and client secret. So maybe just leave line 24 as-is (or if you want, you can link the Base64-encoding procedure to the Client Credential procedure, but don't explicitly tell the user how to do it).

packages/@okta/vuepress-site/docs/guides/implement-grant-type/main/saml2assert/use-flow.md
```diff
@@ -38,3 +38,43 @@ Note the parameters that are being passed:
...
+1. In Windows Explorer, right-click `C:\temp`, and then select **CMD Prompt Here** from the context menu.
+1. Enter the following command to encode the client ID and client secret:
+
+    `copycertutil -encode appCreds.txt appbase64Creds.txt`
+
+1. Locate and open `appbase64Creds.txt` in `C:\temp`, copy its contents, and then close the file.
+
+> **Note:** Delete the `appCreds.txt` and the `appbase64Creds.txt` files after you finish.
+
+### Validate access token
```
vanngo-okta, line 78: I think "Validate access tokens" is already mentioned in the nextsteps.md snippet?

packages/@okta/vuepress-site/docs/guides/implement-grant-type/main/clientcreds/use-flow.md
```diff
@@ -6,7 +6,7 @@ If you are not using existing libraries, you can make a direct request to Okta's
 
 The Client Credentials flow is intended for server-side (confidential) client applications with no end user, which normally describes machine-to-machine communication. Your client application needs to have its client ID and secret stored in a secure manner. You can find the client ID and secret on the **General** tab for your app integration.
 
-[Base64 encode](#base64-encode-the-client-ID-and-client-secret) the client ID and secret and then pass through [Basic Authentication](https://tools.ietf.org/html/rfc7617) in the request to your [Custom Authorization Server's](/docs/concepts/auth-servers/#custom-authorization-server) `/token` endpoint:
+Base64 encode the client ID and secret (as shown below) and then pass through [Basic Authentication](https://tools.ietf.org/html/rfc7617) in the request to your [Custom Authorization Server's](/docs/concepts/auth-servers/#custom-authorization-server) `/token` endpoint:
```
vanngo-okta, line 9:
```suggestion
[Base64 encode the client ID and secret](#base64-encode-the client-id-and-client-secret) and then pass through [Basic Authentication](https://tools.ietf.org/html/rfc7617) in the request to your [Custom Authorization Server's](/docs/concepts/auth-servers/#custom-authorization-server) `/token` endpoint:
```
sorry for this nit-pick: our style guide tells us to avoid using "below" and use a link, or use "later" or "the following"

packages/@okta/vuepress-site/docs/guides/implement-grant-type/main/saml2assert/use-flow.md
```diff
@@ -1,4 +1,4 @@
-Before you can begin this flow, you must collect the SAML assertion from the Identity Provider and make sure that it is [Base64-encoded](https://www.base64decode.org/). You can then use the assertion in the API call to the [Authorization Server's](/docs/concepts/auth-servers/#custom-authorization-server) `/token` endpoint.
+Before you can begin this flow, you must collect the SAML assertion from the Identity Provider and make sure that it is Base64-encoded (as shown below). You can then use the assertion in the API call to the [Authorization Server's](/docs/concepts/auth-servers/#custom-authorization-server) `/token` endpoint.
```
vanngo-okta, line 1:
```suggestion
Before you can begin this flow, you must collect the SAML assertion from the Identity Provider and make sure that it is Base64-encoded. You can then use the assertion in the API call to the [Authorization Server's](/docs/concepts/auth-servers/#custom-authorization-server) `/token` endpoint.
```

vanngo-okta (changes requested):
The base64 encoding procedure for Client Credentials look good!
I'm not sure if the procedure is required for the SAML one. If you want to leave it in, then you probably have to change the Client ID and Client secret bit, because SAML assertions are different.

packages/@okta/vuepress-site/docs/guides/implement-grant-type/main/saml2assert/use-flow.md
chrisdmills-okta: I've just removed this stuff; no worries.

packages/@okta/vuepress-site/docs/guides/implement-grant-type/main/saml2assert/use-flow.md
chrisdmills-okta: Yup, removed.

packages/@okta/vuepress-site/docs/guides/implement-grant-type/main/clientcreds/use-flow.md
chrisdmills-okta: I used this because the link wouldn't work (there seems to be a weird bug with linking to it). I've updated it to "later" to be more correct as per style guide.

vanngo-okta (approved):
Base64 procedure for client credential looks good to me! 
Approving to not hold you back 🚀

packages/@okta/vuepress-site/docs/guides/implement-grant-type/main/saml2assert/use-flow.md
chrisdmills-okta: Yup, removed this one.

PR #2829, opened 2021-11-20
Removing sample code n/a from customize-tokens-dynamic guide

chrisdmills-okta:
Another small bit of cleanup work arising from the single page guide conversion work.


susanharper-okta (approved):
LGTM

PR #2830, opened 2021-11-21
Removing Sample code n/a from customize-authz-server guide nut facts

chrisdmills-okta:
Another small fix to get the nut facts updated to the latest agreed guidelines, this time for this customize-authz-server guide.


susanharper-okta (approved):
LGTM

PR #2831, opened 2021-11-21
Remove Sample code n/a from customize-tokens-static guide nut facts

chrisdmills-okta:
This was recently converted to a single page guide, but it needed a small tweak to update the nut facts to the latest agreed guidelines.


susanharper-okta (approved):
LGTM

PR #2841, opened 2021-11-30
Delete sampleCode.md experiment

chrisdmills-okta:
In this PR I used the delete button in the GitHub web UI to try to get rid of the case insensitive duplicates that we've ended up with.

I've just deleted one so far, in this PR. I'd like to suggest that we see what this looks like, making sure that it gets rid of just the `sampleCode` variant and not the `samplecode` variant, and then if it works, doing the same for the others?

We can't seem to make these changes locally because of case insensitive file systems ;-|


chrisdmills-okta:
Looking at https://github.com/okta/okta-developer-docs/tree/chrisdmills-okta-patch-1/packages/%40okta/vuepress-site/docs/guides/sign-into-web-app/main/aspnet, this seems to do the trick. What do you think @susanharper-okta ?

PR #2884, opened 2021-12-09
Update external Identity Provider documentation

stretch0:
## Description:
- The current documentation makes it seem like a user should build the URL with all the query params themselves. It is not clear that there behind the scene events happening which generate and store PKCE codes. I have added documentation to suggest using the SDK as to help make it clear to other developers.

### Resolves:

* Makes documentation more clear


susanharper-okta:
@closing this and going to update the note higher up in this section to make it more clear about using the SDK.

PR #2956, opened 2022-01-13
fix inconsistent spring boot sample code section

chrisdmills-okta:
## Description:
- **What's changed?** The nutrition facts sample code section on https://developer.okta.com/docs/guides/protect-your-api/springboot/main/ is inconsistent with the sample code for the other platforms — it is too long, and it feels like the cURL stuff should be moved. This PR removes the cURL example completely, and combines the Spring Initializr instructions with the first line. My rationale here is that
  - it is not our responsibility to give our readers several options to create a sample app, just to teach them about using Okta technologies with their (sample) apps. The update still gives them an option that works.
  - The spring boot getting started guide mentions using Spring Initializr in the first line, and the initializr app itself is pretty intuitive, so I think the new placement of our Initializr instructions makes snese in the context of the other linked resources.

### Resolves:

* [OKTA-460424](https://oktainc.atlassian.net/browse/OKTA-460424)


danielmaharry-okta:
Looks good to me

christinechoy-okta (approved):
LGTM!

PR #2980, opened 2022-01-25
change roleId to roleIdOrLabel and specify resources are required whe…

rickyyang-okta:
…n creating resource sets

## Description:
- change roleId in member operations to roleOrLabel since role names can also be used to specify role name as well here
- Change "resources" parameter to be required when creating a resource set

### Resolves:

* [OKTA-462047](https://oktainc.atlassian.net/browse/OKTA-462047)

Checks to make sure the changes in docs reflect public api behaviour + changes to docs

// Creating resource set requires resources
image

image

// Listing members of a binding using role label instead of role id
image

image

// Getting a member for a binding
image

image

// Adding more members to a binding
image

image

// Deleting a member from a binding
Can't show response from DELETE method, but did confirm locally that it worked

image

These cases weren't specified in the ticket, but I found differences between behaviour and what's listed in the docs so included them in here.

// Getting a binding for a role
image

image

// Deleting a binding for a role
Can't show response from DELETE, but can confirm locally that having a role label instead of role id succeeded in deleting the binding

image


brentschaus-okta (approved):
LGTM!

vasishtajayanti-okta (approved):

vasishtajayanti-okta:
Thanks for finding the additional discrepancies!

PR #2983, opened 2022-01-25
Fix okta.yaml sample config file for OIE Embedded SDK go and aspnet guides

danielmaharry-okta:
## Description:
- **What's changed?** 

In the aspnet and go details for the oie-embedded-with-sdk sample, the example yaml config was formatted incorrectly. Using it as written would have resulted in an empty config. I have corrected it from

```
okta:
idx:
  clientid: blah blah
```

to

```
okta:
  idx:
    clientid: blah blah
```

which is now correct

### Resolves:

Not tied to a JIRA ticket


susanharper-okta (approved):
LGTM 🚢

PR #3001, opened 2022-02-02
Oie by default 1 webauthn js

robertnicolo-okta:
## Description:
- WebauthN for Javascript SDK

### Resolves:

* [OKTA-459402](https://oktainc.atlassian.net/browse/OKTA-459402)

Deploy link:

https://61fa99ab0c9964008b14ee42--reverent-murdock-829d24.netlify.app/docs/guides/authenticators-web-authn/nodeexpress/main/


chrisdmills-okta (commented):
Hey @robertnicolo-okta ! This is looking pretty good. I had a few thoughts:

In the "The case for WebAuthN" section:

* "reduce the sign-in friction" -> "reduce sign-in friction"
* It is my understanding that WebAuthn is a browser-based web standard, which is why it effectively provides the local communication you talk about. I'd make it clear that it is a web standard, link to the spec and/or the MDN page (https://developer.mozilla.org/en-US/docs/Web/API/Web_Authentication_API), and also talk about why this local communication mitigates phishing attacks (round trip to the server not needed when WebAuthn can be used, right?)
* You talk about mobile banking apps in your text — do you mean web apps? I thought native apps can't use Web Authn because it needs a browser to function? I might be wrong...and in which case, clarify this, as "*Web* Authn" does imply "needs a web browser". At least, it does to me.
* This is possibly further supported by the fact that Jeff has crossed out the Android and iOS integration guides from the spreadsheet — we don't need to create them. But this further highlights that we need to be clear about what platforms we are suppoing with Web Authn integration, why the ones that aren't supported, are not supported, and what the alternatives would be in those cases.

In "Example authentication flow":

* "security communicate" -> "securely communicate"
* "biometrics authenticator prompts the user for a fingerprint" -> This is a factor, right? Also we should probably generalize a bit more. Maybe "biometrics factor prompts the user for a fingerprint (you could use any factor here as required)"
* "WebauthN" -> "Web authn" . Keep it consistent (I think there's a couple of instances throughout the doc)

In "Update configurations"

* "WebAutha" -> "WebAuthn"
* "FIDO2(WebAuthN)" needs a space? Also "N" to "n"

In "Verify browser version"

* "The Embedded SDK sample app for the Javascript SDK only supports Chrome (version 88 and higher)." -> wow, does this not work in other browsers? Not Firefox? Did you test it in any other browsers? What about mobile browsers versus desktop browsers? Is it supported in mobile and desktop Chrome?
* As well as expanding this info, I'd take the bit about browser support out of a note; it is really important. What do we recommend for users that don't have the required browser support? Do we recommend making multiple factors available so that they'll have one available to use regardless?
* "Javascript" -> "JavaScript"

Typos in the flow diagrams:

* "finger print" (should be "fingerprint"? Make consistent throughout?)
* conset
* Not a typo, but in places where you've got a circular motion that makes a call out and the result goes back to the same actor in the flow (e.g. "Generate challenge for WebAuthn enrollment"), haven't you got the second arrow in each case pointing in the wrong direction? They should be flowing in a semi-circle, not both  pointing towards the action.

Feedback on the different steps

* It's a bit odd that the "Configuration updates" link points to a section called "Update configurations". I'd say you need to change the link to "Update configurations" to keep it consistent and in line with our imperative style. Also, maybe it could be "Update configuration" -> you tend to talk about configuration as one amorphous blob of stuff, rather than multiple items.
* "passing in the webauthn value from IdxOption.value" -> this seems a bit confusing at first, as this object is not shown anywhere, and the value passed to OktaAuth.idx.proceed() is { authenticator }. I'm assuming that's destructuring being used there, which makes the code even harder to read. Showing the IdxOption.value would make it easier to grok.
* "Redirect the user to page" -> missing a "the"
* So the JS object shown in step 4 is the value of IdxTransaction? Would it be better to therefore include `IdxTransaction =` in the code to make this explicit? Or is this a returned object that doesn't get defined in the code anywhere? Sorry if I'm getting this confused ;-)
* Also, the fist block in step 5 seems a bit confusing — the values being passed to JSON.stringify() surely wouldn't work, as the names are not standalone objects, but parts of an object chain? So for example `authenticatorEnrollments` would need to be `IdxTransaction.authenticatorEnrollments`?
* I think the code indentation also doesn't help — the object structure is harder to follow than it should be, but this is easy to fix.
* Some of the above comments also apply to the CredentialCreationOptions object later on.

PR #3025, opened 2022-02-11
Small content design updates

ulyanash:
## Description:
- Updated style for ul, ol, code blocks
- Added markdown definition list:
First Term
: This is the definition of the first term

### Resolves:

* [6.Small content design updates](https://trello.com/c/2fKHFzAh/6-6small-content-design-updates)


chrisdmills-okta (changes requested):
Hi @ulyanash ! Thanks for the updates. 

I'm first going to do a review of the updated numbered bullets. They look pretty much fine, except for the following:

1. The background circle is too dark. You've got the `background-color` set as `#DCDCDC`, whereas in the figma design, they are set as `#D7D7DC` with
`opacity: 0.3;`. In RGBA terms, this would be `rgba(215,215,220,0.3)`, and that's probably the best way to handle this, I'd suspect? Or you could convert to an equivalent rgb color

chrisdmills-okta (changes requested):
@ulyanash next, on to the `<code>` styling. This needs work. It doesn't look like you've implemented the requirement for this one.

https://www.figma.com/file/mHbOK2lNv7AFSViRQVGxUY/Improve-%3Ccode%3E-styling-on-developer.okta.com%2Fdocs?node-id=511%3A419 shows what inline `<code>` elements should be updated to. On the PR, the inline `<code>` blocks look the same as on the live site.

Instead, you seem to have added extra padding to `<pre><code>`, for example on http://localhost:8080/docs/guides/enable-cors/main/#xmlhttprequest. A little bit of extra padding would look good, but this is too much.

chrisdmills-okta (commented):
@ulyanash last, the definition lists. These look really good. The syntax won't work in all cases — for example you can't put multiple definitions on the same term, or multiple terms on the same definition. And you can't put multiple lines of markup source inside the definition term.

But you can include multiple lines as long as you include them all as HTML on one line. And it works pretty well for simple cases, which is much better than we currently have.

Screenshot 2022-02-11 at 14 44 32

The styling looks fine.

The page Ib used to text the syntax was http://localhost:8080/docs/concepts/redirect-vs-embedded/#comparison-between-deployment-models

chrisdmills-okta (approved):
@ulyanash these new changes look perfect, thanks! Merging ;-)

susanharper-okta (commented):
LGTM

PR #3063, opened 2022-02-23
fix(docs): update getUserInfo call in okta-react

cxlucas:
## Description:
- **What's changed?**
- I guess that the function call `getUserInfo()` may be wrong in this documentation.
   - Note: I changed just the _Function-based component_ example, I didn't test if the _Class-based component_ could have the same behavior.

- Following the documentation, it throws an error in React, because the function is not in `oktaAuth`, but in `oktaAuth.token`:
image

- In JS documentation, I found that it could be called as "authClient.token.getUserInfo", and the React Okta seems to have the same pattern, using `oktaAuth.token.getUserInfo()`
  - https://github.com/okta/okta-auth-js/blob/master/README.md

### DIFF
--- a/packages/@okta/vuepress-site/docs/guides/sign-into-spa/main/react/getuserinfo.md
+++ b/packages/@okta/vuepress-site/docs/guides/sign-into-spa/main/react/getuserinfo.md
@@ -19,7 +19,7 @@ const Home = () => {
       // When user isn't authenticated, forget any user info
       setUserInfo(null);
     } else {
-      oktaAuth.getUserInfo().then(info => {
+      oktaAuth.token.getUserInfo().then(info => {
         setUserInfo(info);
       });
     }


vanngo-okta (commented):
Hi @cxlucas, 
Thank you for taking the time to let us know about updating `oktaAuth.getUserInfo()` in the React SPA example. We have recently updated the SPA guides to reflect the updated methods from Auth JS SDK:

* https://developer.okta.com/docs/guides/sign-in-to-spa-embedded-widget/react/main/
* https://developer.okta.com/docs/guides/sign-into-spa-redirect/react/main/

Note: The examples in these guides use `getUser()`, which is an alias for `token.getUserInfo`. See: https://github.com/okta/okta-auth-js#getuser.

The new guides resolves your issue and we're going to proceed to close this PR.

PR #3099, opened 2022-03-01
minor fixes for php web app quickstart

aaronpk:
* use `public` folder since that's the PHP convention and used later down in the tutorial
* fix missing function definition in the redirect block


chrisdmills-okta (approved):
Looking good; thanks @aaronpk !

PR #3124, opened 2022-03-08
update code blog links to tagged page URLs

chrisdmills-okta:
## Description:
- **What's changed?** The blog links at the bottom of the pages under https://developer.okta.com/code/ are somewhat buggy/problematic, as they also show a bunch of github links, and they don't show the pages in much of a useful order. This PR updates those to point to specific tag URLs (see https://developer.okta.com/blog/tags/). Note that I've not yet update the pages for the following, as there are a number of competing tags and it is not clear which one I should use:
  - .NET and .NET Core
  - Node/Node.js
  - Vue/VueJS
- Note: The blog links won't work in localhost, as the build doesn't build the blog (that uses a different system). But they are correct. 

### Resolves:

* [OKTA-476205](https://oktainc.atlassian.net/browse/OKTA-476205)

### DIFF
--- a/packages/@okta/vuepress-site/code/android/index.md
+++ b/packages/@okta/vuepress-site/code/android/index.md
@@ -68,4 +68,4 @@ Other guides:
 * [Validate access tokens](/docs/guides/validate-access-tokens)
 * [Validate ID tokens](/docs/guides/validate-id-tokens)
 
-> **Note**: Browse our [Android Developer Blog posts](/search/#q=android&f:@commonoktasource=[Developer%20blog]) for further useful topics.
+> **Note**: Browse our [Android Developer Blog posts](/blog/tags/android/) for further useful topics.
--- a/packages/@okta/vuepress-site/code/android/index.md
+++ b/packages/@okta/vuepress-site/code/android/index.md
@@ -68,4 +68,4 @@ Other guides:
 * [Validate access tokens](/docs/guides/validate-access-tokens)
 * [Validate ID tokens](/docs/guides/validate-id-tokens)
 
-> **Note**: Browse our [Android Developer Blog posts](/blog/tags/android/) for further useful topics.
+> **Note**: Browse our recent [Android Developer Blog posts](https://developer.okta.com/blog/tags/android/) for further useful topics.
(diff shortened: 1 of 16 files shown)


packages/@okta/vuepress-site/code/android/index.md
```diff
@@ -68,4 +68,4 @@ Other guides:
 * [Validate access tokens](/docs/guides/validate-access-tokens)
 * [Validate ID tokens](/docs/guides/validate-id-tokens)
 
-> **Note**: Browse our [Android Developer Blog posts](/search/#q=android&f:@commonoktasource=[Developer%20blog]) for further useful topics.
+> **Note**: Browse our [Android Developer Blog posts](/blog/tags/android/) for further useful topics.
```
vanngo-okta, line 71:
```suggestion
> **Note**: Browse our [Android Developer Blog posts](https://developer.okta.com/blog/tags/android/) for further useful topics.
```

vanngo-okta: Maybe we should use the full URL so that it opens a new tab for the blogs?

vanngo-okta (commented):
The blog links you have so far look good. Do you want to use full URLs? That way, it opens up a new tab for the blogs. The blog layout looks very different from the dev doc site, so that might make more sense. For the various flavours of the same tag ("dot-net", "dot-net-core", etc.), do you want to add two links? It's not pretty, but at least you get coverage. For "vue", "vuejs" - I don't know. It looks weird to add 2 links for these. Maybe just pick the most used tag with recent blogs?

packages/@okta/vuepress-site/code/android/index.md
mraible: I would suggest saying "Browse our recent [...]" since they're ordered by date descending. Also, it's possible we have posts missing tags, so let us know if you see any that might be mistagged.

chrisdmills-okta:
Thanks for the feedback folks!

@vanngo-okta :

* I've updated the links to use absolute URLs so that they open in a new tab.
* I've chosen a single tag for each of the ones that were outstanding, and I'm chatting to the dev advocate folk about making sure they converge on those tag choices as much as possible (see https://github.com/oktadev/okta-blog/issues/1103).

@mraible :

* I agree with you about the word "recent", and I've added it.

vanngo-okta (approved):
LGTM! 🚀 
@mraible: Can you review and close PR3053 if this PR resolves the Java blog post issue? Thank you!

PR #3131, opened 2022-03-09
Various fixes to Dan Marma feedback

chrisdmills-okta:
## Description:
- **What's changed?** Dan gave us some useful feedback on the recent docs updates for the OIE GTM push. 

## Fixes

The fixes are as follows (I'll add to this as I do more):

1. Make the app type terms (and their order) consistent between https://developer.okta.com/docs/guides/quickstart/main/ and https://developer.okta.com/code/, and the dev.okta homepage. I decided to use the following terms and order:

* Server-side web app/Server-side: In my experience "server-side" is more common than "back-end" when describing a web app, as opposed to a back-end service such as an API.
* Front-end web app/Front-end: Again, in my experience "front-end" is more common than "client-side" when describing an app, plus normally we are talking about SPAs, but not every front-end JS app is an SPA.
* Mobile/native app, or Mobile/native: I think this is fine as-is.


danielmaharry-okta (approved):
lgtm!

PR #3145, opened 2022-03-11
OKTA-459126 Add pagination offset example

vanngo-okta:
## Description:
- **What's changed?** Add pagination offset example in the ASA Intro. Add a link from the ASA Projects API (List Projects for a team operation) to the Pagination example.

### Netlify preview:
* https://622fa341ea05b000a7ad43d4--reverent-murdock-829d24.netlify.app/docs/reference/api/asa/projects/#request-path-parameters

### Resolves:

* [OKTA-459126](https://oktainc.atlassian.net/browse/OKTA-459126)

### DIFF
--- a/packages/@okta/vuepress-site/docs/reference/api/asa/introduction/index.md
+++ b/packages/@okta/vuepress-site/docs/reference/api/asa/introduction/index.md
@@ -51,6 +51,7 @@ ASA provides information about pagination in `Link` headers:
 * A Link header that contains `rel="prev"` indicates that a previous page is
   available and contains the URL that should be used to fetch it.
 
+
 To fetch a full list of resources, Clients should use the following approach
 when retrieving lists:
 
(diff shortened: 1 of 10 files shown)


packages/@okta/vuepress-site/docs/reference/api/asa/introduction/index.md
```diff
@@ -61,6 +62,14 @@ when retrieving lists:
 3. Check for a Link header value with `rel="next"`. If such a value exists,
    fetch it and repeat steps two and three.
 
+For resources that support pagination, specify the `offset` query parameter to fetch the list from a specific page cursor. Use the UUID string of the page cursor from the `Link` header. For example:
```
phoebeyu-okta, line 65: it won't always be a uuid, but whatever is in the `next` Link header should be used here

waltergoulet-okta (commented):
Trying this step out via postman I think we can simplify it to say that the user can simply use the value of the link from the header to fetch the next page. In other words, since we include the full URL in that header it feels like we can skip mentioning the offset field entirely and just direct users to use that full URL instead.

vanngo-okta (commented):
Hi @waltergoulet-okta, can we update the text to this?

> For resources that support pagination, use the URL from the `rel="next"` Link header to fetch the list from a specific page. This URL includes the `offset` query parameter that indicates the specific page cursor for the fetch operation. For example:

I think we need to mention `offset` somewhere since the Projects API contains 2 instances of this parameter, but no description on how to use it. Let me know what you think.
(See existing doc here: https://developer.okta.com/docs/reference/api/asa/introduction/#pagination)

phoebeyu-okta (approved):

waltergoulet-okta:
Looks good to me

waltergoulet-okta (approved):

susanharper-okta (approved):
LGTM

PR #3168, opened 2022-03-18
OKTA-481120: Add requestType, remove eventAction for Telephony hooks

diptidesai-okta:
## Description:
- These changes remove the `eventAction` field used by telephony hooks in BETA and docs the field `requestType` that was newly added by inline hooks framework
- April 2022 release 2022.04.0

### Resolves:

* [OKTA-481120](https://oktainc.atlassian.net/browse/OKTA-481120)


susanharper-okta (approved):
LGTM

PR #3216, opened 2022-04-05
Correcting 2 examples for null and blank

PabloValarezo:
The 2 examples for checking for null and blank attributes are wrong because they use the inequality ("!=") instead of the equality ("==") operators. 
This PR fixes the two examples.

## Description:
- **What's changed?** 
  - The 2 examples for checking for null and blank attributes are wrong because they use the inequality ("!=") instead of the equality ("==") operators. 
  Unsure
### Resolves:


brianduffield-okta:
Thanks for the PR @PabloValarezo . Will get this reviewed and approved.

chenningzhang-okta (approved):

PR #3240, opened 2022-04-12
Upgrade webpack to 4.46.0

francoislehoux-okta:
## Description:
- **What's changed?** Upgrade webpack to 4.46.0

### Resolves:
* [OKTA-481884](https://oktainc.atlassian.net/browse/OKTA-481884)

### DIFF
--- a/packages/@okta/vuepress-site/package.json
+++ b/packages/@okta/vuepress-site/package.json
@@ -42,7 +42,7 @@
     "vuepress-plugin-code-copy": "^1.0.6",
     "vuepress-plugin-robots": "^1.0.1",
     "vuepress-plugin-sitemap": "^2.1.2",
-    "webpack": "^3.0.0",
+    "webpack": "^4.46.0",
     "xml2js": "^0.4.19"
   },
   "browserslist": "> 0.25%"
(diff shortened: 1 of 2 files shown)


TiffanyNg-okta (commented):
I guess there are no breaking changes to be fixed? Do you need to do any of these: https://webpack.js.org/migrate/4/ ?

shawyu-okta:
https://6255f2460e89ed03559a4b05--reverent-murdock-829d24.netlify.app

gaurigattani-okta (approved):

shawyu-okta (approved):
lgtm. i was surprised also that there were no changes we needed to apply for v3 -> v4 migration but looks like the vuepress-cli is already ready for v4 as far as configs go.

i had been mostly concerned about the `mode` flag addition but it appears to be in the [configs already](https://github.com/vuejs/vuepress/blob/38e98634af117f83b6a32c8ff42488d91b66f663/packages/%40vuepress/core/lib/node/webpack/createBaseConfig.js#L33-L40) and I also checked that the `yarn dev` version locally still has development sourcemaps, while the `yarn build` version hosted in the Netlify site does not have the sourcemaps (production mode).

francoislehoux-okta:
`Vuepress` can handle `webpack` 4 at most. `vuepress-next` will work with 5

PR #3241, opened 2022-04-12
removing the in front of Okta Expression Language

susanharper-okta:
## Description:
- **What's changed?** removing the in front of Okta Expression Language and some other odds and ends

### Resolves:

* [OKTA-488837](https://oktainc.atlassian.net/browse/OKTA-488837)


brentschaus-okta (changes requested):
LGTM but I think you missed a few "the"'s (see attached image).

PR #3263, opened 2022-04-19
Upgrade dependencies with available patch and minor upgrades

francoislehoux-okta:
This is a first pass of multiple like this. This will slowly get us back into a state where everything is on the latest version.

## Description:
- **What's changed?** 
- axios
- semver
- include-media
- jshint
- moment
- moment-timezone
- query-string
- vue-select

### Resolves:
* [OKTA-490911](https://oktainc.atlassian.net/browse/OKTA-490911)

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
   "author": "Brian Retterer <brian.retterer@okta.com>",
   "license": "Apache-2.0",
   "devDependencies": {
-    "axios": "^0.21.2",
+    "axios": "^0.26.1",
     "chai": "^4.2.0",
     "chai-as-promised": "*",
     "chalk": "^2.4.2",
--- a/packages/@okta/vuepress-theme-prose/package.json
+++ b/packages/@okta/vuepress-theme-prose/package.json
@@ -25,17 +25,17 @@
   "dependencies": {
     "@okta/odyssey": "^0.0.1",
     "@okta/okta-signin-widget": "^5.4.0",
+    "axios": "^0.26.1",
     "build-url": "^6.0.1",
-    "axios": "^0.21.2",
     "cookies.js": "^2.1.15",
     "css-to-json": "^2.1.0",
-    "include-media": "^1.4.9",
-    "jshint": "^2.12.0",
+    "include-media": "^1.4.10",
+    "jshint": "^2.13.4",
     "jwt-decode": "^3.1.2",
-    "moment": "^2.29.2",
-    "moment-timezone": "^0.5.32",
+    "moment": "^2.29.3",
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -31,7 +31,7 @@
   "author": "Brian Retterer <brian.retterer@okta.com>",
   "license": "Apache-2.0",
   "devDependencies": {
-    "axios": "^0.21.2",
+    "axios": "^0.26.1",
```
shawyu-okta, line 34: I looked over the [changelog](https://github.com/axios/axios/blob/master/CHANGELOG.md) and it seems like the breaking changes do not affect us, but worth grepping through the code and ensuring our uses don't overlap with any breaking changes.

packages/@okta/vuepress-theme-prose/package.json
```diff
@@ -25,17 +25,17 @@
...
     "css-to-json": "^2.1.0",
-    "include-media": "^1.4.9",
-    "jshint": "^2.12.0",
+    "include-media": "^1.4.10",
+    "jshint": "^2.13.4",
     "jwt-decode": "^3.1.2",
-    "moment": "^2.29.2",
-    "moment-timezone": "^0.5.32",
+    "moment": "^2.29.3",
+    "moment-timezone": "^0.5.34",
```
shawyu-okta, line 36: This change is just updating TZ data so we should be good here

packages/@okta/vuepress-theme-prose/package.json
```diff
@@ -25,17 +25,17 @@
...
     "portal-vue": "^2.1.7",
-    "query-string": "^6.13.8",
+    "query-string": "^7.1.1",
```
shawyu-okta, line 38: https://github.com/sindresorhus/query-string/releases shows breaking change to `.stringify` but we don't use it. Should be fine.

shawyu-okta (approved):
I marked any semver changes that could contain breaking changes. I think `axios` might be the only one we need to double check but I do think it's probably also safe.

francoislehoux-okta:
> I marked any semver changes that could contain breaking changes. I think `axios` might be the only one we need to double check but I do think it's probably also safe.

I'll reverify `axios` but I've already done a first compatibility check

francoislehoux-okta:
Did a manual verification. Token retrieval and signup still work: https://625f1508b9cefd219ba8ca12--reverent-murdock-829d24.netlify.app/ @shawyu-okta

shawyu-okta:
Thanks for checking @francoislehoux-okta 

:shipit:

PR #3268, opened 2022-04-20
Upgrade and remove outdated dependencies - 2

francoislehoux-okta:
## Description:
- **What's changed?** 

***Upgraded***
- chai (patch)
- mocha (minor)
- bootstrap (major, we only use `boostrap-grid.css` and there weren't any breaking changes)
- codemirror (minor)
- markdown-it-attrs (minor)
- sass (patch)

***Remove*** (these were dating to 3 years ago and could not find any usage)
- cross-env
- hoek
- joi
- ecstatic
- commander

***Replaced***
serve -> serve-handler (we we're essentially bundling `serve` just to use `serve-handler`)

### Resolves:
* [OKTA-491298](https://oktainc.atlassian.net/browse/OKTA-491298)

### Deploy
https://6260528d9e435210004e9306--reverent-murdock-829d24.netlify.app


gaurigattani-okta (approved):

packages/@okta/vuepress-site/package.json
```diff
@@ -27,16 +27,15 @@
   },
   "dependencies": {
     "@okta/vuepress-plugin-qualtrics-feedback": "^1.0.0",
-    "bootstrap": "^4.5.3",
-    "codemirror": "^5.59.0",
-    "commander": "^2.0.0",
+    "bootstrap": "^5.1.3",
```
shawyu-okta, line 30: looks like bootstrap v5 dropped IE11 support. Just making sure we are OK with this. Not totally sure if it would even have any effect on our site depending on what parts we are using.

shawyu-okta, line 30: looks like bootstrap v5 dropped IE11 support. Just making sure we are OK with this. Not totally sure if it would even have any effect on our site depending on what parts we are using.

package.json
```diff
@@ -32,27 +32,21 @@
...
   },
   "dependencies": {
     "broken-link-checker": "^0.7.8",
-    "hoek": "^6.1.3",
-    "joi": "^14.3.1",
-    "serve": "^11.0.0",
+    "serve-handler": "^6.1.3",
     "vuepress": "1.9.7",
     "vuepress-plugin-chunkload-redirect": "^1.0.3",
     "wait-for-localhost-cli": "^3.0.0"
```
shawyu-okta, line 50: I think most of these can be moved to `devDependencies`, no?

e.g. `serve`/`serve-handler` (and `broken-link-checker`) is used by the broken link checker script

I think the only ones here that might need to stay in `dependencies` is the `vuepress` and `vuepress-plugin-*`

shawyu-okta, line 50: I think most of these can be moved to `devDependencies`, no?

e.g. `serve`/`serve-handler` (and `broken-link-checker`) is used by the broken link checker script

I think the only ones here that might need to stay in `dependencies` is the `vuepress` and `vuepress-plugin-*`

francoislehoux-okta: I'll add a not to the "owner" ticket for to do this in the next pass

packages/@okta/vuepress-site/package.json
francoislehoux-okta: Missed this, I'll move back to latest `4.x.x`

shawyu-okta (approved):
move bootstrap back to 4.x and 🚀

PR #3284, opened 2022-04-25
Archived classic versions of Registration Inline Hook and Set up self-service registration Guides

brianduffield-okta:
## Description:
- **What's changed?** Moved existing Registration Inline Hook and Set up self-service registration guides to the Classic archive; updated navigation; added Classic-only note to archived version of Registration Hook Guide.
- **Is this PR related to a Monolith release?** 2022.05.0

### Resolves:

* [OKTA-491649](https://oktainc.atlassian.net/browse/OKTA-491649)


brianduffield-okta:
I have removed Registration Hook Guide and Self-service registration guide in this PR. Existing xrefs to these docs are causing the failed check.

PR #3297, opened 2022-04-28
Remove ref object from User Types API response.

chenningzhang-okta:
## Description:
- **What's changed?**
Remove ref object from User Types API response; Reorder user type properties in sample response objects.

### Resolves:

* [OKTA-494043](https://oktainc.atlassian.net/browse/OKTA-494043)

@donwoods-okta @daytonwilliams-okta @okta/ud-core

### DIFF
--- a/packages/@okta/vuepress-site/docs/reference/api/user-types/index.md
+++ b/packages/@okta/vuepress-site/docs/reference/api/user-types/index.md
@@ -38,57 +38,49 @@ curl -s -H "Authorization: SSWS ${api_token}" https://${yourOktaDomain}/api/v1/m
 ```json
 [
   {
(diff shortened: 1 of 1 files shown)


donwoods-okta (approved):
LGTM.  Those examples must've been captured using a really early implementation; I cannot even find them as examples in an API review doc.  I think I even recall the uproar and hasty removal of that exposed type name. :trollface:

daytonwilliams-okta (approved):
Goodness.  This sort of internal detail is something we're not really supposed to expose.  Yikes.

`Find in Files` reveals 5 hits for `com.saasure.db.dto.platform.entity.cvd.CVDType`, all five are deleted.  LGTM, :rocket: :sheep: :it:

PR #3376, opened 2022-05-31
OKTA-470845 Minor update to Administrator Roles API reference

brianduffield-okta:
## Description:
- **What's changed?** Updated [supported resources](https://developer.okta.com/docs/reference/api/roles/#supported-resources) table to add URL encoding to one REST URL.

I also have a question about the response example for this particular API. The response example ([All apps of the same type of resource](https://developer.okta.com/docs/reference/api/roles/#all-apps-of-the-same-type-as-resource)) includes a filter and escaped character in the response, but calling the API brings back only individual apps. I believe the given example is incorrect. Can you verify? 

ie: `"https://{yourOktaDomain}/api/v1/apps?filter=name+eq+\"workday\""` , I think, should look like:

`"https://{yourOktaDomain}/home/{targetAppType}/0oa1gjh63g214q0Hq0g4`

(Also, I see the property `_links.apps` but the calls I made return `_links.appLinks`. Do these examples need to be updated?). See this [Google doc](https://docs.google.com/document/d/1c3sOhKZq9HXZnduIrn12oe0alOnC2ElfGnlq6EyeWvY/edit?usp=sharing) for my brief testing.

- **Is this PR related to a Monolith release?** n/a

### Resolves:

* [OKTA-470845](https://oktainc.atlassian.net/browse/OKTA-470845)


susanharper-okta (approved):
LGTM

PR #3460, opened 2022-07-06
Feature/content design update

konstantin-sh-ram:
fix breadcrumbs styles
fix cyan text and bg colors for inline <code>


brianduffield-okta:
@konstantin-sh-ram What is the status of this PR?

PR #3476, opened 2022-07-12
The token management api applies to refresh tokens

danielzeller-okta:
## Description:
- Clarify that the token management APIs act upon refresh tokens.
### Resolves:

* N/A


susanharper-okta (approved):
LGTM

PR #3626, opened 2022-09-16
Aside animation

konstantin-sh-ram:
Create navbar animation like on stripe docs


brianduffield-okta:
Closed as per discussion with team

PR #3662, opened 2022-10-03
removing note

susanharper-okta:
## Description:
- **What's changed?** Removing Okta Classic Engine note - Device auth can be used in both Classic and Identity Engine


IanHakes-okta (approved):
Best update ever!

PR #3676, opened 2022-10-06
Updated monolith URL in the update-error-codes.sh script

brianduffield-okta:
## Description:
- **What's changed?** Updating the `update-error-codes.sh` to use new URL/jar location. This script ran locally and finally updated the `error-codes.json` file.
- **Is this PR related to a Monolith release?** 2022.10.0

### Resolves:

* [OKTA-530544](https://oktainc.atlassian.net/browse/OKTA-530544)


vanngo-okta (approved):
Looks awesome! 👍 Thank you for working on this! 🚀 
One small suggestion:
For the OIDC URL errors, can we change the title of the error table?
Property -> **Error** (`error`)
Description -> **Error Description** (`error_description`)
Or something similar?

scripts/update-error-codes.sh
```diff
@@ -5,7 +5,7 @@
 set -o errexit -o pipefail
 
 # Constants
-readonly url=https://bit.ly/3i5eThn
```
shawyu-okta, line 8: A good improvement here on security! We shouldn't rely on a URL shortener to point us anywhere related to builds.

shawyu-okta (approved):

scripts/update-error-codes.sh
brianduffield-okta: Thanks. I can update the other to scripts that use bitly.

PR #3685, opened 2022-10-10
update sample code with firebase/php-jwt library

aaronpk:
## Description:
- **What's changed?** Updates the PHP API quickstart to the firebase/php-jwt library, since the previous one was deprecated. The sample project is already updated to match https://github.com/okta-samples/okta-php-api-quickstart


chrisdmills-okta (approved):
Changes make sense; thanks @aaronpk

PR #3711, opened 2022-10-21
Update Trusted Origins API to define number of trusted origins returned with and without a limit defined

susanharper-okta:
## Description:
- **What's changed?** Update Trusted Origins API to define number of trusted origins returned with and without a limit defined

### Resolves:

* [OKTA-542264](https://oktainc.atlassian.net/browse/OKTA-542264)


jadsiblini-okta (approved):
LGTM

PR #3776, opened 2022-12-02
OKTA-554436 Update to Expression Language overview

brianduffield-okta:
## Description:
- **What's changed?** Added a note to `Arrays.*` function as per external PR request.
- **Is this PR related to a Monolith release?** n/a

### Resolves:

* [OKTA-554436](https://oktainc.atlassian.net/browse/OKTA-554436)


gabrielsroka:
> CSV files

not "CSV files", just "CSV"

maheshkeralapura-okta:
@nghadge-okta 

Are we sure  the return type of Arrays.flatten()is  an array? 

A lot of the usage of Arrays.flatten() uses the output as a argument to a string function. 

There are a lot more corner cases around array functions, I would hold off updating external documentation before we are sure if these are behavior we want to support or corner cases in how we are leaking evaluation engine internal implementation.

Fwiw - see https://docs.google.com/document/d/1vysg5AazfJfagIg4o8oRwo2XVaWb1TS_SlQHby_dAtA/edit

gabrielsroka:
This is what Okta Support told a customer. I just want it documented.

gabrielsroka:
Example you provided actually uses an array, not a CSV. Please check my original example.

Imho, putting the example below makes it harder to find. Why not put it in the table itself? There's precedence for this in other parts of that doc.

brianduffield-okta:
Apologies, @gabrielsroka, working too quickly. I kept it out of the main table to maintain consistency with the other functions (none of the other functions were repeated or take different input parameters). I'm okay to add it to the table, though. I'll pinging some of the engineering team regarding some of the other contents as well.

PR #3787, opened 2022-12-05
Add new OAuth app link to GitHub social IdP docs

bdemers:
Streamline the process of pointing the developer to the right place, developers can create a new OAuth application by navigating directly to `https://github.com/settings/applications/new`

GitHub enterprise customers with custom domains would likely use a different link, but those users likely would be integrating with SAML or other user store and would not be integrating with Okta this way.


vanngo-okta (approved):
LGTM! I going to get Apps Platform team/John Baldo to review the link before we publish.

johnbaldo-okta (approved):

PR #3800, opened 2022-12-07
Replace moment with luxon

francoislehoux-okta:
## Description:
- Replace `moment` and `moment-timezone` with `luxon`
### Resolves:

* [OKTA-556896](https://oktainc.atlassian.net/browse/OKTA-556896)


shawyu-okta (approved):
lgtm as long as you double-check the output looks the same as before in the preview deploy

francoislehoux-okta:
> lgtm as long as you double-check the output looks the same as before in the preview deploy

Will do. Also made sure the values were matching between `moment` and `luxon` before removing `moment`.

PR #3878, opened 2023-01-12
OKTA-566108 Add rate limit note for resend request in Activate factor section

vanngo-okta:
## Description:
- **What's changed?** Add Note in Factors API for resend after OTP-based activate factor rate limit exceeded
- **Is this PR related to a Monolith release?** Yes 2023.01.1(vuln-after all prod cells)

Screen Shot 2023-01-12 at 11 43 29 AM

### Resolves:

* [OKTA-566108](https://oktainc.atlassian.net/browse/OKTA-566108) for corresponding RN item: OKTA-550739

### DIFF
--- a/packages/@okta/vuepress-site/docs/reference/api/factors/index.md
+++ b/packages/@okta/vuepress-site/docs/reference/api/factors/index.md
@@ -1979,6 +1979,11 @@ The rate limit for a user to activate one of their OTP-based factors (such as SM
 }
 ```
 
+> **Note:** If the user exceeds their OTP-based factor rate limit, then an OTP [resend](#resend-sms-as-part-of-enrollment) request isn't allowed for the same factor. This applies to the following resend endpoints:<br>
+> * `/api/v1/users/${userId}}/factors/${factorId}/resend`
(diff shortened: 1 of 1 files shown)


lalithapidaparthi-okta:
@vanngo-okta - Few comments:

- Do we document the `/me` endpoints? I could not find it. So may be we should remove it from the above note?

- The /api/v1/authn endpoint is for the authentication calls that happen from the SIW through authn api. Since this doc is for the factors API (/api/v1/users/userId/factors/factorId), perhaps we can remove this.

- Do we need explicitly indicate that resend is blocked only for email, sms, and call factors? For others like OV ToTP and Google OTP, there is no resend option from Okta and so this will not apply to them.

vanngo-okta:
@lalithapidaparthi-okta I've updated the note based on your feedback:
* I found some /me endpoints in the Users API, but you're right: the `/me/factors` endpoint isn't documented, so I removed this reference.
* Removed the `/api/v1/authn` endpoint reference.
* Updated the note to explicitly call out the email, call, and sms factors

brentschaus-okta (approved):
LGTM!

PR #3904, opened 2023-01-24
Fix inline prop of shared content

ramgandhi-okta:
## Description:
- **What's changed?** Add deep css selector to `<SharedContent />` component

### Resolves:

* Inline property not working on different environments


ramgandhi-okta:
@mauricesharp-okta FYI/A

PR #4267, opened 2023-07-13
OKTA-627743 Change image div class to appropriate classes

vanngo-okta:
## Description:
- **What's changed?** We were using div="three-quarters" class in some pages, which is not correct. Turns out that the correct CSS is "three-quarter". However, this produced some huge images that were fuzzy. For these scaled images, "original" place-holder class was used which is just no class at all, with no scaling.

### Preview: 
1. https://64b049e3f4c883396931fae1--reverent-murdock-829d24.netlify.app/docs/guides/implement-grant-type/mfaoob/main/#grant-type-flow
2. Preview that uses `three-quarter` for Direct Auth, and `<div>` for no scaling: https://64b6a2366382c60534df6ee2--reverent-murdock-829d24.netlify.app/docs/guides/implement-grant-type/mfaoob/main/#grant-type-flow

### Resolves:

* [OKTA-627743](https://oktainc.atlassian.net/browse/OKTA-627743)

### DIFF
--- a/packages/@okta/vuepress-site/docs/guides/implement-grant-type/main/mfaoob/flow-diagram.md
+++ b/packages/@okta/vuepress-site/docs/guides/implement-grant-type/main/mfaoob/flow-diagram.md
@@ -1,6 +1,6 @@
 ### Direct Authentication MFA OOB flow
 
-<div class="original">
+<div class="three-quarter">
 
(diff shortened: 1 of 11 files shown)


danielmaharry-okta (approved):
Hi @vanngo-okta , can you use `<div>` rather than `<div class="original">` please? Somewhere down the line we'll probably look at the code and wonder how we managed to delete the original class by mistake. 

I should have redrawn the other grant type diagrams to be the same size as the others I've just drawn. Then the scaling will work better. I'll do that separately.

Approving not to hold you up

danielmaharry-okta (approved):
lgtm

PR #4307, opened 2023-08-09
Update index.md adding traditional pkcs flag

derek-uc:
I found I had to use this alternative syntax to convert my private key into a usable format

## Description:
- **What's changed?** New alternative syntax to enable private key auth to work.

### DIFF
--- a/packages/@okta/vuepress-site/docs/guides/terraform-design-access-security/main/index.md
+++ b/packages/@okta/vuepress-site/docs/guides/terraform-design-access-security/main/index.md
@@ -43,6 +43,7 @@ With this flow, Terraform uses credentials to request access to your org. The se
 The credentials used for this flow are a public/private key pair. Okta stores the public key in the service app, and Terraform uses the private key in the configuration. You can [generate the key pair](/docs/guides/implement-oauth-for-okta-serviceapp) in the Okta service app or with your own secure internal methods. See [Create access credentials](/docs/guides/terraform-enable-org-access/main/#create-access-credentials).
 
 > **Note:** The Okta Terraform Provider requires the private key to use PKCS#1 encoding.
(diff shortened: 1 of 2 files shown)


brianduffield-okta:
Thanks for the PR, @derek-uc I'll forward to our engineering teams for confirmation.

mauricesharp-okta:
Hi Derek,

Does the suggested command line refer to converting the key generated using the Okta org Admin Console? I so, it mean that you tried the openSSL command line shown at the end of the section to [create the access credential](https://developer.okta.com/docs/guides/terraform-enable-org-access/main/#create-access-credentials) but it failed, and that the alternate syntax worked?

Thanks

derek-uc:
> Hi Derek,
> 
> Does the suggested command line refer to converting the key generated using the Okta org Admin Console? I so, it mean that you tried the openSSL command line shown at the end of the section to [create the access credential](https://developer.okta.com/docs/guides/terraform-enable-org-access/main/#create-access-credentials) but it failed, and that the alternate syntax worked?
> 
> Thanks

Correct, the I'm assuming a default changed in OpenSSL that no longer converts to pkcs1 but a newer format the `traditional` flag was needed to convert to the older format

derek-uc:
Also, this is only tested on OSX `13.4.1 (22F82)`

mauricesharp-okta:
Hi Derek,

My guess is that it's either the latest release of the openssl tools (3.x) or installed from master. In that case, you can use `openssl-rsa` instead of `openssl` and it should work correctly:

```
openssl rsa -in {ORIGINAL_PRIVATE_KEY} -out {CONVERTED_PRIVATE_KEY}
```

Can you please confirm this works?

Thanks
Maurice

mauricesharp-okta:
The changes for this PR were incorporated into a different ticket and pushed live.
https://oktainc.atlassian.net/browse/OKTA-640695

PR #4357, opened 2023-09-06
Updated information for mobile

mauricesharp-okta:
## Description:
For the three SIW articles:
- Removed the Android and iOS stack selector items
- Added information on where to look for info on mobile

### Resolves:

* [OKTA-643304](https://oktainc.atlassian.net/browse/OKTA-643304)


mauricesharp-okta:
The changes are more complex requiring rewriting some content rather than simply removing the existing mobile entries.
- the current articles blend SIW and IDX
- they give detailed steps for configuration including IDX
- there are a large number of links back to the articles

PR #4364, opened 2023-09-08
[Editing test for Andreea Jurj] Initial editorial pass

aramchek81:
Suggested Improvements
* Title: Needs to be task-driven: What is the Okta Identity Engine *for*? What can the user hope to accomplish?
* Meta description tag: We need to give a little bit more information to guide the user and improve page SEO
* I don't fully understand what this does, but <ApiLifecycle access="ie" /> appears to be a way to tage the topic for identity engine. The tag "ie" could be more explicit. Is this for the blue banner?
* To remove the initial "speed bump" for the reader, it's best to eschew headings such as "overview"; that provide no good content or context. In this case, "How Okta Identity Engine Works" seems to strike a good balance.
* It isn't a good idea to advertise or promise present or future features; documentation isn't marketing materials and such sections should be removed.
* As much as possible, introductory materials need to tie together the basics, so as to help the user to get the necessary information up-front, rather than force the user to "hunt" for the necessary details.
* Colloquialisms such as "over in" must be minimized.
* Structured writing ought to ensure familiar and repeatable, but friendly language as signposts for readers (for instance, "For more information about Foo, see Bar in the Baz guide.")
* A mixture of relative and absolute links can be problematic.
* I wasn't entire sure how admonitions (notes) should be formatted.
* If the build system supports it, Markdown for code blocks should specify the language used to enable syntax highlighting.
* Wherever possible, examples should look real, by using IP addresses or domains as specified in RFC 5737 and RFC 3849. Does ${yourOktaDomain} insert the real example dynamically?
* Is it necessary to document previous behaviour? I greatly reduced this content and placed the remainder within a note, but it could be easily removed.
* When documenting processes, auto-numbering (1. 1. 1. ...) ought to be used.
* There was something wonky with the heading hierarchy in the listing of features, where H3 began to be munged with H2, possibly.
* Documentation strategy changes should be as "thin," minimal, and transparent to the user as possible.
* It's best not to bury important noticwes at the bottom of a doc and to retire or remove them after a set period.
* The fact that there is help.okta.com and developer.okta.com is disorienting and more than a little bewildering. For this reason, it is important to prevent user surprise by indicating after a link on which website they can expect to find the resource.
* Many URLs are not optimized for SEO at all. https://developer.okta.com/code/ makes no sense to me—it ought to be https://developer.okta.com/sdk-tools/
* I saw an inconsistent use of "app" and "application" and erred on the side of using the former to be more conversational.
* I have reorganized the sections for what I hope is a better overall flow for the document, from the highly conceptual to the specific and practical.
* While it's a good idea to use contractions to sound conversational, using the contraction "it's" is problematic for various reasons, including confusion between "it is" and "it has."
* I hate to be overly critical, but the title of the page here is a little...unusual: "Sign users in overview" https://developer.okta.com/docs/guides/sign-in-overview/main/ May I recommend at least "User sign-in overview"?
* In the real-world-scenario, the TOC would be generated automatically
* A thought: How do you handle redirects for renamed pages or pages whose URLs have changed?

## Description:
- **What's changed?** 
I performed a full editorial pass, as instructed by the editing test exercise.
No

### Resolves:

N/A

### DIFF
--- a/packages/@okta/vuepress-site/docs/concepts/oie-intro/index.md
+++ b/packages/@okta/vuepress-site/docs/concepts/oie-intro/index.md
@@ -2,13 +2,13 @@
 title: Generating Dynamic Email by Using the Okta Identity Engine
 meta:
   - name: description
-    content: This section explains how you can use the Okta Identity Engine to customize the style and content of email messages dynamically based on the app that triggers the email notification.
+    content: This page explains the main features of Okta Identity Engine, including the authentication deployment models, the Interaction Code grant type, authentication policies, app context, and app intent links. It also provides SDKs and sample apps that can help you explore these features.
 ---
 
 <ApiLifecycle access="ie" />
 
 # How Okta Identity Engine works
-This page explains the key concepts and features of [Okta Identity Engine](https://help.okta.com/oie/en-us/content/topics/identity-engine/oie-index.htm) and the areas in which it differs from [Okta Classic Engine](/docs/guides/archive-overview/main/).
+This page explains the main features of [Okta Identity Engine](https://help.okta.com/oie/en-us/content/topics/identity-engine/oie-index.htm), including the authentication deployment models, the Interaction Code grant type, authentication policies, app context, and app intent links. It also explains some of the areas in which Okta Identity Engine differs from [Okta Classic Engine](/docs/guides/archive-overview/main/) and provides links to SDKs and sample apps that can help you explore these features.
 
 > **Notes:**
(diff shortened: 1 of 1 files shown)


aramchek81:
(Sorry I didn't create a properly-named branch. I'm just experiencing an intense day at work today, so I'm trying to do a good job while moving fast on this exercise. 😅🙏)

aramchek81:
@andreeajurj-okta Okey dokey. It's all yours. 😅 The full edit pass is in the PR itself and additional comments, annotations, and follow-up questions are in the PR conversation field.

PR #4401, opened 2023-09-21
Updating dark theme "Country/Region" select option style on sign-up page

aidan-d123:
## Description:
When selecting "Country/Region" on the sign-up page (with dark theme enabled), the options are not visible unless hovered over

### Resolves:

Also verified this doesn't change the style of the select box when light theme is enabled

:)


aidan-d123:
@konstantin-sh-ram I'm not able to request reviews but it seems you'd be the right person to review this? Thank you!

emikhasyak-okta (approved):
LGTM! Thank you for providing the fix!

PR #4552, opened 2023-12-11
Reverting url fix

emikhasyak-okta:
## Description:
- Redirection for signup page happens in Amazon configuration, when the page is loaded it already has query parameters dropped from URL

### Resolves:
* [OKTA-671204](https://oktainc.atlassian.net/browse/OKTA-671204)

## Reviewer:
- @paulwallace-okta


paulwallace-okta (approved):
🚢

PR #4607, opened 2024-01-23
OKTA-640287: Add banner component in dev docs

ishan-krishna-okta:
## Description:
- **What's changed?** Added banner component in dev docs. Supports 3 kinds of banners:
  - Info - Type `info`
  - Warning - Type `warning`
  - Danger - Type `danger`

### Resolves:

* [OKTA-640287](https://oktainc.atlassian.net/browse/OKTA-640287)

### DIFF
--- a/packages/@okta/vuepress-theme-prose/global-components/Banner.vue
+++ b/packages/@okta/vuepress-theme-prose/global-components/Banner.vue
@@ -100,23 +100,29 @@
           headerNav = document.getElementsByClassName('header-nav')[0],
           currEle = document.getElementById(this.id),
           fixedHeader = document.getElementsByClassName('fixed-header')[0],
-          allBanners = document.getElementsByClassName('banner');
+          allBanners = document.getElementsByClassName('banner'),
+          sideNavigation = document.getElementsByClassName('on-this-page-navigation');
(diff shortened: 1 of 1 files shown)


ishan-krishna-okta:
Warning Banner
Screenshot 2024-01-23 at 3 40 19 PM


Error Banner
Screenshot 2024-01-23 at 3 40 33 PM


Info Banner
Screenshot 2024-01-23 at 3 41 20 PM

emikhasyak-okta (commented):
The approach currently mixes component logic and CSS and it also adjusts other components layout.
I think it can be designed as follows:
- In banner.vue
  - has template `<div class="banner"> description </div>` (very simplified, you'll )
  - has logic for hiding the banner
  - depending on banner type it will add class to the template like `<div class="banner warning"> description </div>`
- In CSS file
  - figure out styles so that the banner will occupy all horizontal space and use `display:block` or similar so that browser will offset other elements. I hope we can also drop isGlobal setting, that behaviour depends on where we insert the banner.
  - set colors based on banner type

ishan-krishna-okta:
These types of banners are no longer needed. Hence, closing this PR.

PR #4669, opened 2024-02-22
OKTA-682545: Dynamic news feed

ishan-krishna-okta:
## Description:
- **What's changed?** Added support for dynamic news feed in dev docs home page. The feed will be fetched from `developer.okta.com/feed.xml` at runtime.

### Resolves:

* [OKTA-682545](https://oktainc.atlassian.net/browse/OKTA-682545)


packages/@okta/vuepress-theme-prose/components/News.vue
```diff
@@ -77,15 +77,87 @@
...
+  },
+  mounted() {
+    this.getBlogPosts();
+  },
+  methods: {
+    xmlToJs(xmlString) {
+      let xmlDoc = new DOMParser().parseFromString(xmlString, 'text/xml');
+      let result = {};
+
+      function parseNode(node, obj) {
```
emikhasyak-okta, line 112: Wow, nice function and quite a cool solution!

packages/@okta/vuepress-theme-prose/components/News.vue
```diff
@@ -77,15 +77,87 @@
...
+        if (node.nodeType === Node.TEXT_NODE) {
+          if (node.nodeValue.trim() !== '') {
```
emikhasyak-okta, line 114: nit: No strong preference here but if we decide that `obj['#text']` can be an empty string then we can drop this check.

packages/@okta/vuepress-theme-prose/components/News.vue
```diff
@@ -77,15 +77,87 @@
...
+            obj['#text'] = node.nodeValue.trim();
+          }
+        } else if (node.nodeType === Node.ELEMENT_NODE) {
+          let nodeName = node.nodeName;
+          let item = {};
+          if (node.childNodes.length > 0) {
```
emikhasyak-okta, line 120: nit: `if` is not needed. If the length is 0 then `for` loop will exit immediately

packages/@okta/vuepress-theme-prose/components/News.vue
```diff
@@ -20,40 +20,40 @@
...
+            :href="blogPostLinks[0] || 'https://developer.okta.com/blog/2023/04/04/spa-auth-tokens'"
           >
             Read the blog post
           </a>
         </div>
       </article>
       <article class="news__article">
         <p class="news__text dont-break-out">
-          Video
+          {{ blogPostTitles[1] ? 'Blog' : 'Video' }}
```
emikhasyak-okta, line 35: We need to distinguish Video from Blog item and have a different configuration. The current approach will break when we get another Video.

packages/@okta/vuepress-theme-prose/components/News.vue
```diff
@@ -77,15 +77,87 @@
...
 </template>
+
+<script>
+const axios = require('axios');
+
+export default {
+  name: 'News',
+  data() {
+    return {
+      blogPostTitles: [],
```
emikhasyak-okta, line 100: Instead of empty data we can store the current news as default values. i.e.
```
articles: [
  {
     title: 'How Authentication and Authorization Work for SPAs',
     link: 'https://developer.okta.com/blog/2023/04/04/spa-auth-tokens',
     isBlog: true
  },
  {
     title: 'Podcast: Phishing-Resistant Authenticators with Megha Rastogi',
     link: 'https://www.youtube.com/watch?v=PiY5HDp0ABI',
     isBlog: false
  }
  ...
]

```

when we get RSS successfully we will replace articles [0, .. 4]. At that point we can use `v-for` to generate the whole news template for the articles array.

emikhasyak-okta: Maybe an even better approach would be to store JSON file with articles and have a script to update it from RSS. When writers want to get a newer RSS they would run `yarn update-rss` which will fetch XML and update JSON and then just create PR and push the change.
The advantage is we can patch Blog/Video property easily, we can configure a single item manually and we can test new RSS.

emikhasyak-okta (commented):
It's a good progress. I left a bunch of small comments.

I guess the main question is do we want to fetch RSS every time dynamically with risk that data format may change and break what we have on dev docs?
Or do we want an easier way to update RSS as a separate yarn command so that any writer can pull the latest changes and commit those in a more predictable manner?

packages/@okta/vuepress-theme-prose/components/News.vue
ishan-krishna-okta: We will get only blogs in the RSS feed. The video that we see currently in dev docs was added explicitly.

packages/@okta/vuepress-theme-prose/components/News.vue
```diff
@@ -77,15 +77,87 @@
...
+        }
+      }
+
+      parseNode(xmlDoc.documentElement, result);
+
+      return result;
+    },
+    async getBlogPosts() {
+      try {
+        const result = await axios.get('https://developer.okta.com/feed.xml');
```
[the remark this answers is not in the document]
ishan-krishna-okta: Fair point. I have made the requested changes.

packages/@okta/vuepress-theme-prose/components/News.vue
```diff
@@ -20,72 +20,144 @@
...
+  data() {
+    return {
+      blogPosts: initialBlogPosts
+    }
+  },
+  created() {
+    this.BLOG_TYPES = BLOG_TYPES;
+  },
+  mounted() {
+    this.getBlogPosts();
```
emikhasyak-okta, line 101: `getBlogPosts` is async, does it block the `mounted` call?
If we add console log right after HTTP request in `getBlogPosts` and another console log right after `this.getBlogPosts();`
```
async getBlogPosts() {
  await axios.get('https://developer.okta.com/feed.xml'); 
  console.log('recieved response')
}
```
```
mounted() {
    this.getBlogPosts();
    console.log('after getBlogPosts');
}
```

Which message you'll get first? If it is 'after getBlogPosts' then `mounted` function does not wait for `this.getBlogPosts()` and depending on timings we can serve old news or new news for the same user.

emikhasyak-okta (approved):
Great job! 🚢

PR #4743, opened 2024-04-01
[OKTA-711870] Fix cloudfront URL indexing in google search

ishan-krishna-okta:
## Description:
- **What's changed?** To prevent cloudfront URL being indexed in google search, we can do these two things -
   - Add cloudfront URL in robots.txt
   - Add canonical URL to all pages under developer.okta.com pointing to `developer.okta.com/[corresponding-page-route]` so that google/any search engine indexes developer.okta.com instead of cloudfront pages.

### Resolves:

* [OKTA-711870](https://oktainc.atlassian.net/browse/OKTA-711870)


packages/@okta/vuepress-site/.vuepress/config.js
```diff
@@ -314,6 +314,7 @@ module.exports = ctx => ({
               '/docs/reference/api/archive-myaccount/',
               '/docs/reference/api/myaccount-migration/',
               '/docs/reference/csi-delauth-hook/',
+              'd28m3l9ryqsunl.cloudfront.net/',
```
emikhasyak-okta, line 317: Here we have paths listed for dev docs host https://github.com/okta/okta-developer-docs/blob/7f6a8a62d6d20472fdcb88ea4ef640d6b9e3c365/packages/%40okta/vuepress-site/.vuepress/config.js#L290
And it looks like there is no way to provide multiply hosts https://www.npmjs.com/package/vuepress-plugin-robots

packages/@okta/vuepress-site/.vuepress/config.js
```diff
@@ -352,6 +353,8 @@ module.exports = ctx => ({
         $page.redir = `/docs/guides/${found.guide}/${found.sections[0]}/`
       }
     }
+    
+    frontmatter.canonicalUrl = `https://developer.okta.com${path}`;
```
emikhasyak-okta, line 357: This might work, I can't say if it's a good approach.
Also found this on aws:
https://repost.aws/questions/QUlRZYDOhmTm-PEAyS-I6muQ/how-to-avoid-google-to-index-cloudfront-served-content

emikhasyak-okta (commented):
Let's try approach with canonical url and remove CF from robots

packages/@okta/vuepress-site/.vuepress/config.js
ishan-krishna-okta: Yeah I saw this post earlier. I need to discuss this with Doug Chan as it needs some investigation on CF config also. Since he is OOO for a couple of days, I put in this code to help alleviate the problem a bit.

packages/@okta/vuepress-site/.vuepress/config.js
ishan-krishna-okta: Removed it.

PR #4783, opened 2024-04-25
[OKTA-712149]: Moved Netlify manual job to circleci

ishan-krishna-okta:
## Description:
- **What's changed?** Moved Netlify manual job from github actions to circleci.

### Resolves:

* [OKTA-712149](https://oktainc.atlassian.net/browse/OKTA-712149)


.circleci/config.yml
```diff
@@ -54,6 +54,44 @@ jobs:
...
+      - run:
+          name: Has secrets?
+          command: |
+            if [ -z "$NETLIFY_AUTH_TOKEN" ] || [ -z "$NETLIFY_SITE_ID" ]; then
+              echo "Missing required secrets."
+              exit 1
+            fi
+      # Step to cancel workflow
+      - run:
+          name: Cancel workflow if failure
```
emikhasyak-okta, line 72: Is this step required? If any step fails it will cancel the whole job. I'm not sure how this will work and also how to actually get into this step.

.circleci/config.yml
```diff
@@ -92,5 +130,5 @@ workflows:
   netlify_manual:
     when: << pipeline.parameters.trigger_netlify_manual >>
```
emikhasyak-okta, line 131: Will writers need to enter trigger_netlify_manual when they trigger a preview build?
If so could we simplify it to simply `preview`?

emikhasyak-okta (commented):
Overall looks good! Have a couple small questions in comments.
Could you also attach a link to CCI preview build?

.circleci/config.yml
ishan-krishna-okta: Its a precautionary measure to ensure that the necessary secrets (NETLIFY_AUTH_TOKEN and NETLIFY_SITE_ID) are provided before proceeding with the job.

While it's true that any step failure will cancel the entire job, having this explicit check ensures that the job fails early if essential secrets are missing, providing a more informative error message than if the job fails due to a step relying on those secrets.

The step is not strictly required for the job to function, but it adds an extra layer of safety and clarity to our configuration. We can remove the step also.

For now, I'll remove this step as the build will itself fail if the secrets are not there.

.circleci/config.yml
ishan-krishna-okta: Good suggestion. Changed it to `preview`.

ishan-krishna-okta:
Here is the circleCI job link for `netlify_manual` - https://app.circleci.com/pipelines/github/okta/okta-developer-docs/2489/workflows/a6f1d7c5-f66e-44c9-bf1d-bb7642e17ad2/jobs/4297

.circleci/config.yml
ishan-krishna-okta: Yes, this isn't required. Removed it.

emikhasyak-okta (approved):
Great job! 🚀

PR #4951, opened 2024-07-10
chore(ci): Reduce resource class size of two pipeline jobs

adamharvey-okta:
## Description:

Looking at the resource usage for the build checks job, the CPU seems like it could be tuned down, and while RAM does at one point spike between 1/3 to just below 1/2 usage of 16 GB on XLarge, the job itself seems to set a Node limit of roughly 4GB, far below the 8GB which is on a Large resource class. So thought it would be wise to try to see if we can tune this down for improved resource usage.

Same seems to apply to nightly, so made that change too.

Sample run from a recent mainline build:

image

https://app.circleci.com/pipelines/github/okta/okta-developer-docs/3586/workflows/23e11945-8f10-4cb5-bb2b-cc49f254eb81/jobs/7256/steps

### Resolves:

* n/a

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -35,7 +35,7 @@ jobs:
   nightly_link_checker:
     docker:
       - image: cimg/node:14.21.3-browsers
-    resource_class: xlarge
+    resource_class: large
     environment:
       NODE_OPTIONS: '--max_old_space_size=3584'
     steps:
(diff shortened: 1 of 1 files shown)


adamharvey-okta:
Seems like it worked?  Pegs CPU for a short bit, but the duration was no longer than normal (in fact, this run happened to be slightly shorter at 11mins than others I've seen).

image
https://app.circleci.com/pipelines/github/okta/okta-developer-docs/3729/workflows/35f6b510-dc36-4b13-ae56-f0e0bc9dd1ce/jobs/7723/resources

emikhasyak-okta (approved):
LGTM! 🚢 
I doubt we need `xlarge`, in the worst case the build should run slower

PR #4971, opened 2024-07-19
chore(ci): Yarn caching enhancements and pipeline tweaks

adamharvey-okta:
## Description:

This PR provides a number of enhancements to the CI/CD pipeline in the repo.

* Implement Yarn dependency caching.
  * Currently, each Yarn install step seems to take roughly 35 seconds.
  * By caching the dependencies into CircleCI, and restoring them, the restore + install collecitvely seems to take on average 21 seconds. 
  * This is *unscientifically* a 40% reduction in that step. 😉
  * Practically, it just shaves a little over 10 seconds of a job which takes ~10 minutes.
* Use CircleCI Test features.
  *  The Cypress reporter for JUnit was added, as well as the pipeline collecting those values.
  * CircleCI will now be able to calculate test metrics, and more easily raise their values/results into the UI.
  * Overtime, this'll allow the project to make use of features like automated flaky test detection, trends, etc.
  * Right now, Cypress only runs two tests, technically, so this might be a direction to start to enhance that or break out individual steps of one test into multiple tests for test clarity.
* Reduce Netlify resource class size.
  * This job seems to have a bit more room than it needs, so downgraded it from an XLarge to a Large.
  * image
* Fixed Artifactory reference in `yarn.lock`.
  * I believe I noticed some delays in downloads, and if I ran a `yarn install` locally, the lock file for this updated.
  * Believe this should be correct, as external CCI open source cloud runners would not have access to that Artifactory, so it probably adds delay we don't need or warnings, and would mean potential issues for open source contributors installing the dependencies on their local. Correct me if I'm missing an aspect here!
* Fixed repo name in `package.json`.

### Resolves:

* [OKTA-750775](https://oktainc.atlassian.net/browse/OKTA-750775)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,11 +2,31 @@ version: 2.1
 
 orbs:
   general-platform-helpers: okta/general-platform-helpers@1.8
-  
+
 parameters:
   preview:
     type: string
     default: ""
+
+commands:
+  install-deps:
+    steps:
+      - restore_cache:
+          name: Restore Yarn dependency cache
+          keys:
+            # Find a cache corresponding to this specific yarn.lock checksum
+            # when this file is changed, this key will fail (correctly)
+            - v1-yarn-deps-{{ checksum "yarn.lock" }}
+            # Find the most recently generated cache used from any branch
+            - v1-yarn-deps-
+      - run:
+          name: Install dependencies
+          command: yarn install --frozen-lockfile
+      - save_cache:
+          name: Save Yarn dependency Cache
+          key: v1-yarn-deps-{{ checksum "yarn.lock" }}
+          paths:
+            - node_modules/
 
 jobs:
   inclusive-linting:
@@ -25,9 +45,7 @@ jobs:
         NODE_OPTIONS: '--max_old_space_size=3584'
       steps:
         - checkout
-        - run:
-            name: Install dependencies
-            command: yarn install --frozen-lockfile
+        - install-deps
         - run:
             name: Lint
             command: yarn lint-all
@@ -46,9 +64,7 @@ jobs:
       NODE_OPTIONS: '--max_old_space_size=3584'
     steps:
       - checkout
-      - run:
-          name: Install dependencies
-          command: yarn install --frozen-lockfile
+      - install-deps
       - run:
           name: Run script and capture logs
           command: yarn build
@@ -98,10 +114,7 @@ jobs:
               exit 1
             fi
       - checkout
-      - run:
-          name: Install dependencies
-          command: |
-            yarn install --frozen-lockfile
+      - install-deps
       - run:
           name: Build preview
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,31 +2,12 @@ version: 2.1
 
 orbs:
   general-platform-helpers: okta/general-platform-helpers@1.8
+  node: circleci/node@5.2
 
 parameters:
   preview:
     type: string
     default: ""
-
-commands:
-  install-deps:
-    steps:
-      - restore_cache:
-          name: Restore Yarn dependency cache
-          keys:
-            # Find a cache corresponding to this specific yarn.lock checksum
-            # when this file is changed, this key will fail (correctly)
-            - v1-yarn-deps-{{ checksum "yarn.lock" }}
-            # Find the most recently generated cache used from any branch
-            - v1-yarn-deps-
-      - run:
-          name: Install dependencies
-          command: yarn install --frozen-lockfile
-      - save_cache:
-          name: Save Yarn dependency Cache
-          key: v1-yarn-deps-{{ checksum "yarn.lock" }}
-          paths:
-            - node_modules/
 
 jobs:
   inclusive-linting:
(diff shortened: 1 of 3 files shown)


adamharvey-okta:
May have to fix my unverified commits, and do a local squash so as not to add the debug noise to this PR. Still in draft!

yarn.lock
```diff
@@ -10345,7 +10345,7 @@ qs@6.9.7:
 
 qs@~6.5.2:
   version "6.5.3"
-  resolved "https://registry.yarnpkg.com/qs/-/qs-6.5.3.tgz#3aeeffc91967ef6e35c0e488ef46fb296ab76aad"
+  resolved "https://artifacts.aue1e.internal:443/artifactory/api/npm/npm-okta-master/qs/-/qs-6.5.3.tgz#3aeeffc91967ef6e35c0e488ef46fb296ab76aad"
```
adamharvey-okta, line 10348: fix to an Artifactory URL not accessible externally

package.json
```diff
@@ -34,7 +34,7 @@
     "eslint": "eslint \"packages/@okta/vuepress-theme-prose/**/*.vue\"",
     "lint-all": "yarn eslint --max-warnings=0 && yarn stylelint"
   },
-  "repository": "git@github.com:okta/okta.github.io",
+  "repository": "git@github.com:okta/okta-developer-docs",
```
adamharvey-okta, line 37: fix of repository name

package.json
```diff
@@ -5,7 +5,7 @@
   ],
   "scripts": {
     "dev": "yarn workspace @okta/vuepress-site dev",
-    "build": "yarn lint-all && yarn workspace @okta/vuepress-site build",
+    "build": "yarn workspace @okta/vuepress-site build",
```
adamharvey-okta, line 8: separation of these, so that a build doesn't ALWAYS fire a lint.  if this is too impacting to other development processes, happy to put back.

.circleci/config.yml
```diff
@@ -82,7 +109,7 @@ jobs:
   netlify_manual:
     docker:
       - image: cimg/node:14.21.3-browsers
-    resource_class: xlarge
+    resource_class: large
```
emikhasyak-okta, line 112: We actually need `xlarge` otherwise preview build fails with out of memory error

yarn.lock
emikhasyak-okta: That's strange I can download file from `https://registry.yarnpkg.com/qs/-/qs-6.5.3.tgz#3aeeffc91967ef6e35c0e488ef46fb296ab76aad` it should be available.
If we change dependency to artifactory that will break the build for everyone outside of Okta VPN.

.circleci/config.yml
adamharvey-okta: Yah. When looking at the resources for a manual run, it did go green but it was peaking CPU and nearing peaks for Mem, so since this one is less frequently used it's likely much better to keep it at XLarge. Reverting.

package.json
emikhasyak-okta: Yes that makes sense. If we remove lint from build command we need to run lint check before [yarn build](https://github.com/okta/okta-developer-docs/blob/master/scripts/publish.sh#L24) in production part.
```
if ! yarn lint-all; then
    echo "Lint check failed"
    exit ${BUILD_FAILURE}
fi
```

yarn.lock
adamharvey-okta: oh this is odd; no the intent was to REMOVE the artifactory URL and REPLACE with the yarnpkg.com one.  maybe something else is up.... looking....

package.json
adamharvey-okta: agree, believe i've already done that, but through a CCI step?  like so?

https://github.com/okta/okta-developer-docs/blob/4373ecf9668429dc76198ad20df9882882a8c8da/.circleci/config.yml#L50-L55

emikhasyak-okta: Okay let's leave it.
It's either we run the linting two times then builds on CCI and Bacon will have all the checks but we're wasting resources. Or we run everything on CCI and for Bacon we know that all checks are green already.

adamharvey-okta: Sorry @emikhasyak-okta I was missing context, but re-reading I think I'm groking. You're saying because it's a package setting, we're actually running that build step in Bacon ALSO, and so we were properly linting in Bacon. And since i separated it out here in CCI, it's now NOT linting in Bacot - which could be good (no duplicate resourcing) or bad (not good coverage?)

Alternatively, could do this same lint separation to a new step in Bacon if we think useful?

otherwise if this is just confusing happy to revert! hoping i'm understanding the root query here.

yarn.lock
adamharvey-okta: yeah very odd. every time i run yarn locally, it reverts to the artifactory link. i'd guess it's doing this in CI also? wonder if it does it on your local. regardless, reverted it to keep whats in the repo...

package.json
emikhasyak-okta: Yes, we're running the same build command in bacon from a shell script. Let's keep your change as it is, since we already linting everything in CCI.

emikhasyak-okta (approved):
Great job! 🚢

PR #5187, opened 2024-11-08
OKTA-824464: Removed if-env package which was causing corejs vuln

ishan-krishna-okta:
## Description:
- **What's changed?** core-js vuln was being caused by `if-env` package. We can fix this by removing the package and adding alternate code where if-env is used.

### Resolves:

* [OKTA-824464](https://oktainc.atlassian.net/browse/OKTA-824464)


emikhasyak-okta (approved):
Looks good! 🚢

PR #6283, opened 2026-07-07
AuthJS SDK: Sign in to SPA - update React examples

brentschaus-okta:
#### What's changed?
Updates the React examples in [Sign in to your SPA with Auth JS](https://developer.okta.com/docs/guides/sign-in-to-spa-authjs/react/main/), which had gone stale against current @okta/okta-auth-js (8.x) and were reported by a customer as "incorrect and badly out-of-date."

**`basic-sign-in.md`**
- Rewritten around Auth JS 8.x's Step Mode default: `idx.start()` begins the transaction, and `idx.proceed({ step: 'identify', ...inputValues })` names the remediation step explicitly instead of letting the client guess.
- Added a "Start the sign-in transaction" section showing the `idx.start()` bootstrap step — the previous version jumped straight to handling submission with no documented way to get the initial form fields.
- Added a "The full sign-in component code" section with a complete `App.jsx` so readers see the whole flow together.
- Shows actual usage of the `Security` component from `@okta/okta-react`, including a real `restoreOriginalUri` implementation (`history.replace` + `toRelativeUrl`) — previously listed as a required dependency but never shown in any code sample.
- Switched `config.js` from `process.env` to Vite's `import.meta.env` convention, matching the guide's Vite-based scaffold instructions; added `pkce: true`.
- Fixed a template literal bug in `REDIRECT_URI` (missing `$`).
- Dropped the unused `hasErrorInUrl` import.
- Commented out the sequence diagram — it still shows the old single-call `idx.authenticate(username, password)` flow and no longer matches the two-call `idx.start()` → `idx.proceed({ step })` flow. Design request filed to update it: [OKTA-1220704](https://oktainc.atlassian.net/browse/OKTA-1220704).

**`download-sample.md`**
- Reframed the cloned sample as environment/org verification only — it still hardcodes Legacy Mode internally, so its source no longer matches this guide's Step Mode code.
- Dropped the stale `USE_INTERACTION_CODE` env var (removed from the SDK).
- Switched scaffold instructions from Create React App to Vite, and documented the required `.env`/`VITE_` prefix setup.
- Expanded the `react-router-dom` v5 note into a clearer known-limitation callout.

**`sign-in-form.md`**
- Fixed `apps.js`/`formtransformer.js` naming typos.
- Switched to `createRoot` (React 18) from the deprecated `ReactDOM.render`.

Scope is intentionally limited to the React tab — Vue and Angular tabs are untouched and may need a similar pass separately.

#### Verification
The Step Mode call shapes (`idx.start()`, `step: 'identify'`, `restoreOriginalUri`) were reviewed by [@jaredperreault-okta](https://github.com/jaredperreault-okta) as SDK owner/SME across two review passes ([first](https://github.com/okta/okta-developer-docs/pull/6283#discussion_r3536876012), [second](https://github.com/okta/okta-developer-docs/pull/6283#discussion_r3552382779)), but this hasn't been smoke-tested against a live dev org yet — flagging for a final pass before merge.

Is this PR related to a Monolith release? No.

### Resolves:

[OKTA-889035](https://oktainc.atlassian.net/browse/OKTA-889035)

### Netlify Preview Link:

[Preview link](https://tbs-okta-889035-spa-authjs-react-update--dev-docs-preview.netlify.app/docs/guides/sign-into-spa-redirect/react/main/)

### DIFF
--- a/packages/@okta/vuepress-site/docs/guides/sign-in-to-spa-authjs/main/react/basic-sign-in.md
+++ b/packages/@okta/vuepress-site/docs/guides/sign-in-to-spa-authjs/main/react/basic-sign-in.md
@@ -20,18 +20,11 @@ export default {
   issuer: ISSUER,
   redirectUri: REDIRECT_URI,
   scopes: ['openid', 'offline_access', 'profile', 'email'],
-  pkce: true,
-  idx: {
-    // Okta Auth JS 8.x defaults idx.proceed() to Step Mode, which requires every
-    // call to name a step or action. This guide uses the older, generic
-    // remediation-driven pattern below instead (calling idx.proceed() with just
-    // the form values), so Legacy Mode is enabled here to keep that working.
-    enableLegacyMode: true
-  }
+  pkce: true
 };
 ```
 
-> **Note:** Legacy Mode is on a deprecation path in Okta Auth JS but is still fully supported. It's what lets the generic, response-driven form pattern in this guide work without hardcoding every remediation step by name. See [Step Mode vs Legacy Mode](https://github.com/okta/okta-auth-js/blob/master/docs/idx.md#step-mode-vs-legacy-mode) in the Auth JS SDK docs if you want to build against the newer, explicit-step approach instead.
+> **Note:** Okta Auth JS 8.x requires every `idx.proceed()` call to name the remediation step it's submitting. This guide uses that Step Mode pattern throughout. See [Step Mode vs Legacy Mode](https://github.com/okta/okta-auth-js/blob/master/docs/idx.md#step-mode-vs-legacy-mode) in the Auth JS SDK docs if you're migrating an older app that relies on the deprecated, generic remediation-driven pattern.
 
 ### Instantiate the Okta Auth JS client
 
@@ -91,15 +84,67 @@ Pass `transaction.nextStep` into `formTransformer` (see [Basic sign-in flow](#ba
 
 ### Handle the password authentication
 
-Review the `app.js` file for details on handling a successful password authentication by receiving the `SUCCESS` status and storing the returned tokens:
+Name the step you're submitting on every `idx.proceed()` call — `transaction.nextStep.name` holds the value, and it's the same step the form already rendered against. For this password-only use case, the step name is `identify`. Review the `app.js` file for details on handling a successful password authentication by receiving the `SUCCESS` status and storing the returned tokens:
 
 ```JavaScript
-....
-
 const handleSubmit = async e => {
+  e.preventDefault();
+
+  const newTransaction = await oktaAuth.idx.proceed({ step: 'identify', ...inputValues }); // inputValues = username, password
+  console.log('Transaction:', newTransaction);
+
+  setInputValues({});
+  if (newTransaction.status === IdxStatus.SUCCESS) {
+    oktaAuth.tokenManager.setTokens(newTransaction.tokens);
+  } else {
+    setTransaction(newTransaction);
+  }
+};
+```
+
+### The full sign-in component code
+
+With the pieces above in place, here's the complete `App.jsx` for the password-only sign-in flow, with everything shown together:
+
+```JavaScript
+import { useEffect, useState } from 'react';
+import { useHistory } from 'react-router-dom';
+import { OktaAuth, IdxStatus, urlParamsToObject } from '@okta/okta-auth-js';
+import { Security } from '@okta/okta-react';
+import { formTransformer } from './formTransformer';
+import oidcConfig from './config';
+import './App.css';
+
+function createOktaAuthInstance() {
+  const { state } = urlParamsToObject(window.location.search);
+  return new OktaAuth(Object.assign({}, oidcConfig, {
+    state
+  }));
+}
+
+const oktaAuth = createOktaAuthInstance();
+const restoreOriginalUri = () => {};
+
+function SignInForm() {
+  const [transaction, setTransaction] = useState(null);
+  const [inputValues, setInputValues] = useState({});
+
+  useEffect(() => {
+    const startTransaction = async () => {
+      const newTransaction = await oktaAuth.idx.authenticate();
+      setTransaction(newTransaction);
+    };
+    startTransaction();
+  }, []);
+
+  const handleChange = ({ target: { name, value } }) => {
+    setInputValues({ ...inputValues, [name]: value });
+  };
+
+  const handleSubmit = async e => {
     e.preventDefault();
 
-    const newTransaction = await oktaAuth.idx.proceed(inputValues); // inputValues = username, password
+    const newTransaction = await oktaAuth.idx.proceed({ step: 'identify', ...inputValues });
     console.log('Transaction:', newTransaction);
 
     setInputValues({});
@@ -110,5 +155,40 @@ const handleSubmit = async e => {
     }
   };
 
-  ...
-  ```
+  if (!transaction?.nextStep) {
+    return null;
+  }
+
+  const { inputs } = formTransformer(transaction.nextStep)({});
+
+  return (
+    <form onSubmit={handleSubmit}>
+      {inputs.map(({ label, name, type, required }) => (
+        <label key={name}>
+          {label}
+          <input
+            name={name}
+            type={type}
+            required={required}
+            value={inputValues[name] || ''}
+            onChange={handleChange}
+          />
+        </label>
+      ))}
+      <button type="submit">Sign in</button>
+    </form>
+  );
+}
+
+function App() {
+  const history = useHistory();
+
+  return (
+    <Security oktaAuth={oktaAuth} restoreOriginalUri={restoreOriginalUri}>
+      <SignInForm />
+    </Security>
+  );
+}
+
+export default App;
+```
--- a/packages/@okta/vuepress-site/docs/guides/sign-in-to-spa-authjs/main/react/basic-sign-in.md
+++ b/packages/@okta/vuepress-site/docs/guides/sign-in-to-spa-authjs/main/react/basic-sign-in.md
@@ -3,7 +3,7 @@ Review the simple password-only sign-in use case from the sample app.
 <!-- The sequence diagram below is out of date with the Step Mode rework in this
 guide (it shows a single idx.authenticate(username,password) call, not the
 two-call authenticate()-then-proceed({ step }) flow). Commented out until the
-updated diagram is ready — see OKTA-1220702 for the design request tracking
+updated diagram is ready — see the design request ticket (TBD) for tracking
 this.
 <div class="full">
 
--- a/packages/@okta/vuepress-site/docs/guides/sign-in-to-spa-authjs/main/react/basic-sign-in.md
+++ b/packages/@okta/vuepress-site/docs/guides/sign-in-to-spa-authjs/main/react/basic-sign-in.md
@@ -2,9 +2,9 @@ Review the simple password-only sign-in use case from the sample app.
 
 <!-- The sequence diagram below is out of date with the Step Mode rework in this
 guide (it shows a single idx.authenticate(username,password) call, not the
-two-call authenticate()-then-proceed({ step }) flow). Commented out until the
-updated diagram is ready — see OKTA-1220704 for the design request tracking
-this.
+multi-call proceed({ step }) flow, which primes then submits each remediation
+step in turn). Commented out until the updated diagram is ready — see
+OKTA-1220704 for the design request tracking this.
 <div class="full">
 
 ![Sequence diagram that displays the interactions between the resource owner, SDK, authorization server, and resource server for a basic SPA password sign-in flow.](/img/oie-embedded-sdk/password-only-spa-authjs-flow.svg)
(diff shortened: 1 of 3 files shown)


packages/@okta/vuepress-site/docs/guides/sign-in-to-spa-authjs/main/react/basic-sign-in.md
```diff
@@ -8,30 +8,40 @@ Review the simple password-only sign-in use case from the sample app. This use c
...
   issuer: ISSUER,
   redirectUri: REDIRECT_URI,
   scopes: ['openid', 'offline_access', 'profile', 'email'],
+  pkce: true,
+  idx: {
+    // Okta Auth JS 8.x defaults idx.proceed() to Step Mode, which requires every
+    // call to name a step or action. This guide uses the older, generic
+    // remediation-driven pattern below instead (calling idx.proceed() with just
+    // the form values), so Legacy Mode is enabled here to keep that working.
+    enableLegacyMode: true
```
jaredperreault-okta: `enableLegacyMode` is supposed to be an opt-out option to ease migrating from Authjs 7.x to 8.x. It should _not_ be the recommended path.

The pattern of the having the client "guess" what remediation you want to proceed with is buggy and unpredictable. This change was purposely made, so developers explicitly choose which remediation to proceed with

packages/@okta/vuepress-site/docs/guides/sign-in-to-spa-authjs/main/react/basic-sign-in.md
```diff
@@ -8,30 +8,40 @@ Review the simple password-only sign-in use case from the sample app. This use c
...
+  }
 };
 ```
 
+> **Note:** Legacy Mode is on a deprecation path in Okta Auth JS but is still fully supported. It's what lets the generic, response-driven form pattern in this guide work without hardcoding every remediation step by name. See [Step Mode vs Legacy Mode](https://github.com/okta/okta-auth-js/blob/master/docs/idx.md#step-mode-vs-legacy-mode) in the Auth JS SDK docs if you want to build against the newer, explicit-step approach instead.
```
jaredperreault-okta: > but is still fully supported.

That is not entirely accurate. If a bug is reported against a legacy mode flow it's VERY unlikely we will fix it. Therefore I don't think we can claim legacy mode is fully supported

packages/@okta/vuepress-site/docs/guides/sign-in-to-spa-authjs/main/react/basic-sign-in.md
brentschaus-okta: @jaredperreault-okta Good catch on both — you're right that we shouldn't be steering people toward Legacy Mode as the recommended path. Here's what we're planning instead:

- Drop `idx: { enableLegacyMode: true }` from the config entirely and remove the "still fully supported" note.
- Rewrite `handleSubmit` to call `idx.proceed({ step: 'identify', username, password })` instead of the generic `idx.proceed(inputValues)` — naming the step explicitly instead of letting the client guess.
- We're going with `step: 'identify'` for this password-only case (single sign-on policy factor), based on how `authenticate.ts` in the SDK's own test suite maps a username/password submit to the `identify` remediation. We'll smoke-test this against a real dev org before merging, but wanted to flag the exact call we're planning in case you see a problem with it up front.
- One heads up: `samples/generated/react-embedded-auth-with-sdk` (the sample this guide tells readers to clone) still hardcodes `enableLegacyMode: true` and the generic `idx.proceed(inputValues)` pattern throughout (`config.js`, `App.jsx`, `GeneralForm.jsx`). We're not going to rewrite the guide to match that sample — we'll de-emphasize it as a code reference (keep it just for environment setup) and treat the guide's own snippets as the source of truth. Flagging in case that sample is also on your radar to update separately.

Does this match what you had in mind, or is there anything about the `identify` step assumption you'd correct before we go write it?

jaredperreault-okta: Yea it does

brentschaus-okta: Pushed the rework — `basic-sign-in.md` now drops `enableLegacyMode` entirely and `handleSubmit` calls `oktaAuth.idx.proceed({ step: 'identify', ...inputValues })`.

@jaredperreault-okta one thing I couldn't do myself: smoke-test that exact call against a live dev org (no org/credentials in my environment). Since you own the SDK behavior this depends on, could you verify as SME that `step: 'identify'` is correct for a password-only sign-on policy, and that the field shape (`username`/`password` spread into the call) matches what the `identify` remediation expects? Want to confirm before this merges with an unverified claim.

brentschaus-okta:
Pushed one more change: commented out the sequence diagram in `basic-sign-in.md`. It still shows a single `idx.authenticate(username,password)` call, which doesn't match the guide's actual two-call flow (`idx.authenticate()` to start, then `idx.proceed({ step: 'identify', ... })` to submit credentials) — and now that the code explicitly names the step, the mismatch is more visible. Filed [OKTA-1220702](https://oktainc.atlassian.net/browse/OKTA-1220702) with InfoDev to get an updated diagram; will re-enable it once that's ready (may land after this PR merges).

brentschaus-okta:
Correction to my earlier comment: OKTA-1220702 was created in error (I created it directly via API instead of cloning the InfoDev UX Design Request template as required) and is being removed. A new design request ticket will be filed properly via the template — will link it here once it exists.

brentschaus-okta:
Design request filed properly (cloned from the InfoDev template): [OKTA-1220704](https://oktainc.atlassian.net/browse/OKTA-1220704). Updated the code comment in `basic-sign-in.md` to reference it, and started the FigJam concept: https://www.figma.com/board/whddo6SnC7Eiy44CkxPFja/OKTA-1220704

packages/@okta/vuepress-site/docs/guides/sign-in-to-spa-authjs/main/react/basic-sign-in.md
```diff
@@ -44,21 +53,104 @@ function createOktaAuthInstance() {
...
-Review the `apps.js` file for details on handling a successful password authentication by receiving the `SUCCESS` status and storing the returned tokens:
+Before you can render a sign-in form, you need an in-progress IDX transaction to drive it. Start one when your component mounts by calling `idx.authenticate()` with no arguments — the response's `nextStep.inputs` tells you which fields to render. For this password-only use case, that's `username` and `password`:
 
 ```JavaScript
-....
+const [transaction, setTransaction] = useState(null);
+
+useEffect(() => {
+  const startTransaction = async () => {
+    const newTransaction = await oktaAuth.idx.authenticate();
```
jaredperreault-okta:
```suggestion
    const newTransaction = await oktaAuth.idx.start();
```

packages/@okta/vuepress-site/docs/guides/sign-in-to-spa-authjs/main/react/basic-sign-in.md
```diff
@@ -69,5 +161,40 @@ const handleSubmit = async e => {
...
+        </label>
+      ))}
+      <button type="submit">Sign in</button>
+    </form>
+  );
+}
+
+function App() {
+  const history = useHistory();
+
```
jaredperreault-okta, line 189:
```suggestion
const restoreOriginalUri = async (_oktaAuth, originalUri) => {
    history.replace(toRelativeUrl(originalUri || '/', window.location.origin));
  };
```

packages/@okta/vuepress-site/docs/guides/sign-in-to-spa-authjs/main/react/basic-sign-in.md
```diff
@@ -44,21 +53,104 @@ function createOktaAuthInstance() {
...
+function createOktaAuthInstance() {
+  const { state } = urlParamsToObject(window.location.search);
+  return new OktaAuth(Object.assign({}, oidcConfig, {
+    state
+  }));
+}
+
+const oktaAuth = createOktaAuthInstance();
+const restoreOriginalUri = () => {};
```
jaredperreault-okta:
```suggestion

```

packages/@okta/vuepress-site/docs/guides/sign-in-to-spa-authjs/main/react/basic-sign-in.md
```diff
@@ -44,21 +53,104 @@ function createOktaAuthInstance() {
 }
 
 const oktaAuth = createOktaAuthInstance();
+const restoreOriginalUri = () => {};
 
 ...
+
+function App() {
+  const history = useHistory();
+
```
jaredperreault-okta, line 57:
```suggestion
const restoreOriginalUri = async (_oktaAuth, originalUri) => {
    history.replace(toRelativeUrl(originalUri || '/', window.location.origin));
  };
```

packages/@okta/vuepress-site/docs/guides/sign-in-to-spa-authjs/main/react/basic-sign-in.md
```diff
@@ -44,21 +53,104 @@ function createOktaAuthInstance() {
 }
 
 const oktaAuth = createOktaAuthInstance();
+const restoreOriginalUri = () => {};
```
jaredperreault-okta:
```suggestion

```

brentschaus-okta:
## Live-org smoke test results

Ran the `basic-sign-in.md` React flow end-to-end against a real Identity Engine preview org (dedicated scratch app + test user, isolated from the PR 6288 test setup — both marked "safe to delete after" for cleanup). This surfaced two real bugs beyond the suggestions Jared already flagged inline (both now resolved as of the latest commit):

1. **`idx.start()` never populates `transaction.nextStep` in Step Mode.** The doc's "Start the sign-in transaction" section had `formTransformer(transaction.nextStep)` running right after `idx.start()`, but `nextStep` only gets computed once you explicitly name a step. That line would throw (`Cannot read properties of undefined`) on mount as written.
2. **The password-only flow is not one combined `idx.proceed({ step: 'identify', username, password })` call.** Against the live org, `identify` only accepts `username`. Submitting it returns a `challenge-authenticator` step that separately requires `credentials`/`password` for the actual password check. This matches Okta's identifier-first sign-in UX (username screen, then password screen) — it's not an org-config quirk, it's how Identity Engine's IDX API works.

Fixed both by reworking the flow around a **prime-then-submit** pattern:
- `idx.proceed({ step })` with no other values resolves that step's renderable fields (this is what makes `nextStep.inputs` show clean field names like `username`/`password` instead of the raw IDX names like `identifier`).
- `idx.proceed({ step, ...inputValues })` submits the real values and reveals the name of the next step to prime.

This still names every step explicitly on every call (nothing is auto-guessed), so it keeps the Step Mode principle from the discussion below intact — it just accounts for password-only sign-in actually taking two round trips through `handleSubmit`, not one.

Verified end-to-end: `idx.start()` → prime `identify` → submit `username` → prime `challenge-authenticator` → submit `password` → `IdxStatus.SUCCESS` with real `accessToken`/`idToken`.

@jaredperreault-okta flagging in case there's an org-config reason a password-only policy might ever resolve `identify` directly to `SUCCESS` in one call rather than always chaining to `challenge-authenticator` — happy to adjust if so, but this was consistent across everything I tried on the test org.

brentschaus-okta:
Re-checked all of @jaredperreault-okta's earlier inline comments against the current diff:

- **`enableLegacyMode` shouldn't be the recommended path** ([r3536876012](https://github.com/okta/okta-developer-docs/pull/6283#discussion_r3536876012)) — dropped entirely, config no longer sets it.
- **"still fully supported" language for Legacy Mode was inaccurate** ([r3536884944](https://github.com/okta/okta-developer-docs/pull/6283#discussion_r3536884944)) — that note is removed.
- **`idx.start()` naming suggestion** ([r3552382779](https://github.com/okta/okta-developer-docs/pull/6283#discussion_r3552382779)) — applied (`idx.authenticate()` → `idx.start()` in both the snippet and the full `App.jsx`).
- **Real `restoreOriginalUri` implementation, not the `() => {}` stub** ([r3552417953](https://github.com/okta/okta-developer-docs/pull/6283#discussion_r3552417953), [r3552424706](https://github.com/okta/okta-developer-docs/pull/6283#discussion_r3552424706)) — applied in both places (`app.js` snippet and full `App.jsx`), using `history.replace(toRelativeUrl(...))`.
- **Blank-line cleanup suggestions** ([r3552422113](https://github.com/okta/okta-developer-docs/pull/6283#discussion_r3552422113), [r3552426246](https://github.com/okta/okta-developer-docs/pull/6283#discussion_r3552426246)) — applied.

All six threads are resolved. The two bugs in the comment above were found during live-org verification on top of these — the `idx.start()` naming fix was correct as suggested, but the surrounding call shape (`step: 'identify'` bundling both `username` and `password` in one submit) still needed the prime-then-submit rework to actually reach `SUCCESS`.

packages/@okta/vuepress-site/docs/guides/sign-in-to-spa-authjs/main/react/basic-sign-in.md
```diff
@@ -46,28 +55,158 @@ function createOktaAuthInstance() {
...
+import { useEffect, useState } from 'react';
+import { useHistory } from 'react-router-dom';
+import { OktaAuth, IdxStatus, urlParamsToObject, toRelativeUrl } from '@okta/okta-auth-js';
+import { Security } from '@okta/okta-react';
+import { formTransformer } from './formTransformer';
+import oidcConfig from './config';
+import './App.css';
+
+function createOktaAuthInstance() {
+  const { state } = urlParamsToObject(window.location.search);
```
jaredperreault-okta: What is the purpose of the `state` param?

brentschaus-okta: Checked this against the live org (SDK source + a live-org test run). `state` here is used by `TransactionManager`'s shared-storage mechanism: if a `state` value is present, the SDK loads/saves the in-progress transaction keyed by that value, so a redirect away and back (social IdP sign-in, email magic links) can reconnect to the transaction it left off. This guide's password-only flow never redirects away from the page, and its own scope note already excludes those redirect callbacks. I ran the full flow against a live org without ever setting `state`, and it reached `SUCCESS` identically every time.

So this `urlParamsToObject`/`state` code in `createOktaAuthInstance` is leftover from the sample app and doesn't do anything for this guide's flow. Removing it now.

jaredperreault-okta (approved):

grahamsmith-okta (approved):

PR #6439, opened 2026-09-14
Show stack selector on oie-upgrade-api-sdk-to-oie-sdk guide

brentschaus-okta:
## Description:
- **What's changed?** Every `<StackSelector>` in `oie-upgrade-api-sdk-to-oie-sdk/main/index.md` had the `noSelector` attribute set, which hides the framework dropdown. Landing on this guide with no framework in the URL (the actual top-nav link) silently defaulted to the Android version of the content with no way to switch stacks. Removed `noSelector` from all 10 instances in this guide's `main/index.md` so the dropdown renders, matching the sibling guide `oie-upgrade-add-sdk-to-your-app`. Scoped to this one guide only.

### Resolves:

* [OKTA-1274923](https://oktainc.atlassian.net/browse/OKTA-1274923)

### Vercel Preview Link:

🤖 Generated with [Claude Code](https://claude.com/claude-code)


brentschaus-okta:
Reverting the content change in this PR: it was based on an incorrect diagnosis.

The visible stack selector dropdown is injected once per page by a layout-level component gated by a `$page.hasStackContent` flag, which is only set for the framework-qualified URLs (`.../android/main/`, etc.) generated in `.vuepress/scripts/build-guides-info.js`. It's unrelated to the `noSelector` attribute on the in-content `<StackSelector>` snippets in this guide's `main/index.md`. Removing it there didn't add the missing selector, and instead broke rendering by turning 10 snippet blocks into 10 independent, out-of-sync mini-selectors.

The bare `/main/` URL (no framework segment) is a secondary, non-canonical route that's excluded from the sitemap on purpose (see the OKTA-745577 comment in `config.js`) and isn't what the site's own nav/sidebar links generate — those always resolve to a framework-qualified URL. A real fix belongs in that shared routing/build logic, not in this guide's content, and would affect every multi-framework guide site-wide, so it needs to be scoped separately.

Marking as draft pending that follow-up.

